PR #9, opened 2013-06-26
Add a few minor usability improvements.

mattdeboard:
This patch improves the usability of ``wordcloud.py`` at the command line by using ``argparse``. I added several switches but of note:

1. ``--skip``: Skip any missing/unfound files specified at command line

2. ``--output-name``: User-specified output filename

3. ``--fontfile``: Specify font file at command line.

I replaced PIL with Pillow which is a drop-in replacement of PIL that is setuptools compatible, and plays much nicer with virtualenvs.

Also, added support for specifying stopwords at the command line. I included ``stopwords.txt`` which is Solr-Lucene's default English stopword list. It's considerably shorter than ``sklearn``'s stopword list.

Finally, added requirements.txt. The ``scipy`` and ``scikit-learn`` requirements will still require some package-fu (i.e. installing necessary libs via the OS package manager) but now at least all requirements are documented and easily installed via ``pip``.


amueller:
outdated, there is now a command line interface.

PR #25, opened 2014-09-29
Fontconfig

jjduhamel:
This uses the python fontconfig bindings to handle system fonts.  This is standard on Linux and can be made to work with OSX pretty easily.


amueller:
I think this was solved more simply by adding a font to the package distribution.

PR #53, opened 2015-03-29
added support for passing custom word weights

sathappanspm:
Updated WordCloud.generate function to take custom (word, weight) pairs

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -301,15 +301,19 @@ class WordCloud(object):
         return words
 
     def generate(self, text):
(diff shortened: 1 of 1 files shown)


amueller:
Don't you think adding a new function would be more discoverable?

amueller:
Thanks for the PR btw :)

amueller:
Wait, I don't understand. Why not just call ``WordCloud.fit_words(text)``?

amueller:
I added a ``from_frequencies`` function (which is just a renaming of ``fit_words`` which returns self). See if that fits your need.

PR #54, opened 2015-04-05
Besides raw text string, tokens can be passed to the 'generate' method

xiaohan2012:
As far as I understood, custom tokenization is not easy to use for the current version. This pull request allows custom tokenization to be used.

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -233,8 +233,8 @@ class WordCloud(object):
         self.layout_ = list(zip(words, font_sizes, positions, orientations, colors))
         return self.layout_
 
-    def process_text(self, text):
-        """Splits a long text into words, eliminates the stopwords.
+    def _get_words(self, text):
+        """Splits a long text into words
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


wordcloud/wordcloud.py
```diff
@@ -233,8 +233,8 @@ def fit_words(self, words):
         self.layout_ = list(zip(words, font_sizes, positions, orientations, colors))
         return self.layout_
 
-    def process_text(self, text):
-        """Splits a long text into words, eliminates the stopwords.
+    def _get_words(self, text):
```
amueller: Why did you remove this public function? Backward incompatible changes make it hard for others to use the code.

wordcloud/wordcloud.py
```diff
@@ -300,8 +318,15 @@ def process_text(self, text):
 
         return words
 
-    def generate(self, text):
-        """Generate wordcloud from text.
+    def generate(self, stuff):
```
amueller: Stuff is not a good parameter name.

amueller: I would rather have different functions for different input formats, like ``generate_from_tokens`` or something like this. It is also not really clear from the docstring what input format is expected.

amueller:
Your code is not python3 compatible (see travis).

wordcloud/wordcloud.py
xiaohan2012: Thanks for pointing out. I don't know that until you told me, as I am only modifying to my own need.

wordcloud/wordcloud.py
xiaohan2012: Good to know that!

amueller:
Sorry for the slow reply. So is the idea here that you could overload specific methods to get custom tokenization?
I think I would rather pass a callable tokenizer to the constructor (or maybe to the function?).
Also, where do you use isstr?

amueller:
this can be achieved using ``generate_from_frequencies``. also, no replies :-/

PR #73, opened 2015-06-12
Colors for word_cloud

evanmiltenburg:
This modification adds different color palettes (in RGB format) to be used for the generation of word clouds. It is very easy to add other palettes: just add a list of RGB tuples to the palettes dict.


evanmiltenburg:
I don't understand the Travis logs: they don't say what's wrong with the script, just `no such option: --use-mirrors`. Something wrong with the way the test is set up, perhaps?

amueller:
hum, maybe there is. I have to check that

evanmiltenburg:
Either way, I'd be happy to improve the code if you like the idea :)
[Here's an album of word clouds](http://imgur.com/a/iY0YK) using the `constitution.txt` file.
Code used to generate the album:

    with open('constitution.txt') as f:
        text = f.read()

    wordcloud = WordCloud(font_path="/Users/Emiel/Library/Fonts/FiraMono-Regular.ttf",
                          background_color="white", width=1024, height=768,
                          color_func=random_color_func, palette=0).generate(text)

    for i in range(0,7):
        wordcloud.recolor(palette=i)
        wordcloud.to_file('constitution' +str(i)+ '.png')

amueller:
If you rebase the travis error will go away. Though I'm not sure about your approach. first off, I don't think I would add a new parameter to the interface. Why don't you just provide a new color_func?

evanmiltenburg:
Thanks for the response. Providing a new color_func will be easy :) I'll open a new pull request once I have the time to implement everything (and add some more color schemes).

Why I didn't write a new color_func straight away:
* I added a new parameter to the interface because that was the only way to pass the preferred color scheme to the color function. Not adding it would force the user to first generate a randomly colored word cloud, and then recolor the word cloud using their preferred color scheme. Conceptually, that seemed a bit odd to me to color the word cloud twice.
* I modified the random color function because the color palettes provide basically the same functionality (randomly color each word), just with a more specific selection of colors.

I'm fine with adding a new function though.

amueller:
I don't understand your first point. There is a constructor parameter for a color_func so they don't have to call ``recolor`` to get their preferred scheme...

evanmiltenburg:
Well, suppose I want to have color palette number 5, but for another word cloud I want to use color palette 6. How else do I pass `5` or `6` to the `palette_color_func`? Adding a `palette` keyword to `WordCloud` seemed like the only way.

    wordcloud = WordCloud(font_path="/Users/Emiel/Library/Fonts/FiraMono-Regular.ttf",
                          color_func=palette_color_func, ????)

amueller:
``color_func = partial(palette_color_func, palette=5)`` ?

evanmiltenburg:
Sure, but that looks a bit inconsistent to me, considering you *do* have random_state. Keeping everything consistent, all `color_func` arguments should be passed using `partial`. I.e. `random_state` and `mask` should be removed from `WordCloud`. Since that breaks backwards compatibility, that's probably not the way to go for you. But then why not add a `palette` argument?

amueller:
``random_state`` is also used in the layout.

Why not add palette? Because it clutters the interface and more people might want to add more color_funcs, and the way you wrote the code will break other people's code where they already provide a custom ``color_func`` that doesn't take a ``palette`` argument.

evanmiltenburg:
Ah, sorry! I understand now. Back to the drawing board :) 

(I hope you do like the colors though.)

amueller:
yeah, they are much nicer. Mine are horrible ^^

evanmiltenburg:
Ok, rewrote the code and the build passes. Are you happy with the way it is structured now? Then I'll try to add some tests.

amueller:
Thanks. I have to think about the structure. Definitely put it in it's own file.
I'm wondering if it would make more sense to have the class be a callable to that it could be passed directly as a ``color_func`` without using ``partial``, by constructing an object containing only one pallet, not a container of many pallets. but I'm not entirely sure.

evanmiltenburg:
Yeah so the question is really what we want the default situation to be like. I think it is really nice to include multiple color palettes by default. But then how to store those? Intuitively, for me, a dict-like structure makes sense. But there might also be other options. I just can't think of one.

If we go this route, then it might be good to *only* expose `palettes` (the instance) and not `Palettes` (the class) to the user. Else it would be confusing to first-time users.

amueller:
I agree with your points. I'm not sure what the way to go is, and I'm not sure I really have time to think about it atm :-/

PR #85, opened 2015-07-23
allow for phrases to be treated as a single word

Fuzzwah:
This fork adds a tiny hack which allows words to be grouped into small 
phrases. To do this the word source is pre-parsed to insert underscores
in place of spaces in the phrases you wish to have treated as a single 
word. The 1 line of code which is changed results in these underscores
being removed when the phrase is written into the word cloud.


amueller:
You could to this by just providing your own tokenization and use the ``generate_from_frequency`` function. Then you wouldn't have to hack anything.

amueller:
check out this example: https://github.com/amueller/word_cloud/blob/master/examples/bigrams.py
I think detection collocations is the way to go. This should probably be ported to the main class and extended to trigrams. any help welcome.

PR #91, opened 2015-08-18
Frequency scaling

amueller:
Finally actually taking the relative frequencies into account.

### DIFF
--- a/examples/simple.py
+++ b/examples/simple.py
@@ -17,4 +17,10 @@ wordcloud = WordCloud().generate(text)
 # Open a plot of the generated image.
 plt.imshow(wordcloud)
 plt.axis("off")
+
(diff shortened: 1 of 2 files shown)


amueller:
just using the ratios (i.e. ``relative_scaling=1``) doesn't look good because it doesn't fill the space. I played around with several non-linear functions of the ratio, and doing ``(new_freq / old_freq) ** (1. / some_number)`` looks similar. But just having a linear function seems more easy to understand.

pcrutchl:
Great, thanks. Trying out now instead of my hacky version.

amueller:
well this is also pretty hacky, but I am not sure what else I could be doing. Play around with ``relative_scaling`` around ``.5`` and ``max_font_size``.

amueller:
Feel free to post pictures ;)

Issue #98, opened 2015-11-05
PIL is deprecated, replace with Pillow


maxupp:
For use under Linux, PIL should be replaced with Pillow, the successor.

amueller:
You mean in the requirements?

maxupp:
Exactly, if possible without rewriting the entire tool ;)

amueller:
It works with either.

amueller:
Pillow provides Image, just as PIL did.

maxupp:
Yes, but wordcloud still requires PIL:
/usr/lib/python3.4/site-packages/wordcloud/wordcloud.py in <module>()
     14 from operator import itemgetter
     15 
---> 16 from PIL import Image
     17 from PIL import ImageDraw
     18 from PIL import ImageFont

ImportError: No module named 'PIL'

amueller:
hm ok.
And you have pillow installed and that doesn't work?

maxupp:
Yep.

amueller:
That is odd, pillow contains the PIL module:
https://github.com/python-pillow/Pillow

You can try changing it to pillow, but I don't think that'll work.

amueller:
Pillow is imported as PIL:
https://pillow.readthedocs.org/en/3.0.x/handbook/tutorial.html

adalmieres:
Hi, there is an announcement at http://pillow.readthedocs.org/en/3.0.x/installation.html
Pillow >= 2.1.0 no longer supports “import _imaging”. Please use “from PIL.Image import core as _imaging” instead.

amueller:
I'm not importing ``_imaging``, right? But I guess I shouldn't import ``Image`` either.
Pull request to use ``from PIL import Image`` welcome.

jasonrhaas:
I'm running into this on Ubuntu 14.04 LTS.

`ImportError: No module named PIL`

Has anyone figured out a fix for this?  I am installing in a clean virtualenv and get that error.

amueller:
Install PIL or Pillow.

amueller:
setup.py uses pillow now. closing.

Issue #96, opened 2015-10-09
PIL setfont() deprecated


methodds:
Hi,

when using word_cloud on windows with python 3.5 (Anaconda), plotting a text works perfectly fine but I always receive a warning from Anaconda refering to PIL: 
"C:\Anaconda3\lib\site-packages\PIL\ImageDraw.py:104: UserWarning: setfont() is deprecated. Please set the attribute directly instead.
  "Please set the attribute directly instead.")"

Is there another way to set the font without receiving the warning? This is the code I used:

```python
wordcloud = WordCloud( font_path='C:\Windows\Fonts\calibrib.ttf', max_font_size=40,
                      width = 800, height=600, stopwords=stopwords_eng).generate(tweet_txts)
```

Thanks for you work!

amueller:
well they say to set the attribute directly ;) PR welcome.

amueller:
fixed in #106

PR #106, opened 2015-12-05
Update font API & merge upper and lower case words when stemming

pendragon-:
Using PIL imports in the examples, just like the main library file for consistency & compatibility.
The bigger change is moving the plural word merges so as to use the lower case words to check for existing words. This results in 'Lukes' being merged together with 'LUKE', for example.

Also fixes the following issues:
- https://github.com/amueller/word_cloud/issues/96
- https://github.com/amueller/word_cloud/issues/98

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -289,9 +289,8 @@ class WordCloud(object):
                     orientation = Image.ROTATE_90
                 transposed_font = ImageFont.TransposedFont(font,
                                                            orientation=orientation)
-                draw.setfont(transposed_font)
                 # get size of resulting text
-                box_size = draw.textsize(word)
+                box_size = draw.textsize(word, font=transposed_font)
                 # find possible places using integral image:
                 result = occupancy.sample_position(box_size[1] + self.margin,
                                                    box_size[0] + self.margin,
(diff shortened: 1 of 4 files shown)


pendragon-:
I am not sure what that I broke anything there:
> ImportError: libgfortran.so.1: cannot open shared object file: No such file or directory
I'd appreciate any pointers on how to fix this.

wordcloud/wordcloud.py
```diff
@@ -361,31 +359,33 @@ def process_text(self, text):
                 continue
 
             # Look in lowercase dict.
-            if word_lower in d:
+            try:
```
amueller: no strong opinion on that but why not...

amueller:
I triggered the test again, it seemed unrelated. Other than that, looks good to me. Thanks!

amueller:
It is this issue: https://github.com/conda/conda/issues/444

I just pushed a fix to master. If master builds successfully (maybe I should have done a PR ^^) can you rebase?

amueller:
seems to work. please rebase and let me know if the tests pass.

pendragon-:
All done & green.

amueller:
thanks :)

stonebig:
could we have a small 1.2.1 to get the fix on Pypi ?

amueller:
should be up now @stonebig

stonebig:
Thanks, Andreas !

PR #107, opened 2015-12-05
downgrade numpy req 

amueller:
so that we can use the travis ubuntu version instead of pip-installing it.


amueller:
works, merging. Should make the tests faster.

Issue #109, opened 2015-12-07
Overlapping words with spaces in "words"


CodeBrauer:
I've added the support, that words can contain space like in https://github.com/amueller/word_cloud/pull/85 but there seems to a problem.

The spaces are not correctly used and the words overlap. - Only on not monospaced fonts.

**Without spaces** *(LoremIpsum)*
![result](https://cloud.githubusercontent.com/assets/2059754/11627084/413b92de-9ce8-11e5-99a4-4785988ac4d4.png)

**With spaces** *(Lorem_Ipsum)*
![result-1](https://cloud.githubusercontent.com/assets/2059754/11627085/414f6bf6-9ce8-11e5-8bbe-f161f3982662.png)

- Font used: ["Bebas Neue"](http://www.fontsquirrel.com/fonts/bebas-NEUE)
- [Wordlist used](https://gist.githubusercontent.com/CodeBrauer/a93c14b8dd9d663cf2a8/raw/844e59d6cf526157bbc98e41744e2668bc9365a1/loremwc.txt)

My modifications in [wordcloud.py](https://github.com/amueller/word_cloud/blob/9aaf533788b1611bfd8b991e89f511d37ee7a886/wordcloud/wordcloud.py):
```
draw.setfont(transposed_font)
pos = (position[1] * self.scale, position[0] * self.scale)
- draw.text(pos, word, fill=color)
+  #draw.text(pos, word, fill=color)
+  draw.text(pos, word.replace('_',' '), fill=color)
return img
```

CodeBrauer:
I tempfixed it using another regex selector:
```
for word in re.findall(r".*", text):
```

amueller:
you can use "from_frequencies" instead and don't need to change any code ;). But then you don't get the tokenization at all. It would be good to make the regex a constructor parameter. PR welcome!

CodeBrauer:
I'm not the python guy, but I can try it :+1:

amueller:
good luck :) shouldn't be too hard!

amueller:
check out this example: https://github.com/amueller/word_cloud/blob/master/examples/bigrams.py

PR #111, opened 2015-12-08
Add a command line interface

pendragon-:
Usable with
```bash
python -m wordcloud.tool -h
```
Let me know what you think.

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -350,7 +350,7 @@ class WordCloud(object):
         d = {}
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
                  else 0)
-        for word in re.findall(r"\w[\w']*", text, flags=flags):
+        for word in re.findall(r"\w[\w']+", text, flags=flags):
             if word.isdigit():
                 continue
 
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -10,10 +10,12 @@ from random import Random
 import os
 import re
 import sys
+import colorsys
 import numpy as np
 from operator import itemgetter
 
 from PIL import Image
+from PIL import ImageColor
 from PIL import ImageDraw
 from PIL import ImageFont
 
@@ -76,6 +78,37 @@ def random_color_func(word=None, font_size=None, position=None,
     if random_state is None:
         random_state = Random()
     return "hsl(%d, 80%%, 50%%)" % random_state.randint(0, 255)
+
+def get_single_color_func(color):
+    """Create a color function which returns a single hue and saturation with.
+    different values (HSV). Accepted values are color strings as usable by PIL/Pillow.
+
+    >>> color_func1 = get_single_color_func('deepskyblue')
+    >>> color_func2 = get_single_color_func('#00b4d2')
+    """
+    old_r, old_g, old_b = ImageColor.getrgb(color)
+    rgb_max = 255.
+    h, s, v = colorsys.rgb_to_hsv(old_r/rgb_max, old_g/rgb_max, old_b/rgb_max)
+    def single_color_func(word=None, font_size=None, position=None,
+                          orientation=None, font_path=None, random_state=None):
+        """Random color generation.
+
+        Additional coloring method. It picks a random value with hue and
+        saturation based on the color given to the generating function.
+
+        Parameters
+        ----------
+        word, font_size, position, orientation  : ignored.
+
+        random_state : random.Random object or None, (default=None)
+          If a random object is given, this is used for generating random numbers.
+
+        """
+        if random_state is None:
+            random_state = Random()
+        r, g, b = colorsys.hsv_to_rgb(h, s, random_state.uniform(0.2, 1))
+        return 'rgb({:.0f}, {:.0f}, {:.0f})'.format(r * rgb_max, g * rgb_max, b * rgb_max)
+    return single_color_func
 
 
 class WordCloud(object):
--- a/wordcloud/wordcloud_cli.py
+++ b/wordcloud/wordcloud_cli.py
@@ -1,0 +1,83 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+r"""Command-line tool to generate word clouds
+Usage::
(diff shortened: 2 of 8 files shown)


wordcloud/wordcloud.py
```diff
@@ -350,7 +350,7 @@ def process_text(self, text):
         d = {}
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
                  else 0)
-        for word in re.findall(r"\w[\w']*", text, flags=flags):
+        for word in re.findall(r"\w[\w']+", text, flags=flags):
```
amueller: actually, I'm not sure about this change any more. Maybe we should just customize the regexp as discussed in #109. Yours would be a better default, though.

amueller:
This looks cool, but I don't have time to review right now. Maybe rename the file to ``command_line`` or something?

amueller:
also, having a callable file might be nicer than doing ``-m``?

pendragon-:
I wanted to have it close to other python callables (think json.tool). If you want to have it callable, you have to call the file exactly. As the module has to be installed, I do not think that's quite as convenient.

amueller:
well you can install files into the bin path, too. pip doesn't only install libraries, it also installs "programs". I wouldn't call flake8 or sphinx using python -m.

pendragon-:
I've tried:
```bash
  File ".../.virtualenvs/wordcloud/lib/python2.7/site-packages/wordcloud/tool.py", line 8, in <module>
    import wordcloud as wc
  File ".../.virtualenvs/wordcloud/lib/python2.7/site-packages/wordcloud/wordcloud.py", line 20, in <module>
    from .query_integral_image import query_integral_image
ValueError: Attempted relative import in non-package
```
as far as I've found, to fix that I'd need to do some path hacking.

amueller:
not sure I follow. the executable should be installed in the bin folder, not the lib folder.

pendragon-:
Yeah right, I just noticed too. I'll check again.

amueller:
sorry, I've been gone for a bit. Is this ready from your perspective?

wordcloud/wordcloud_cli.py
```diff
@@ -0,0 +1,104 @@
...
+                          orientation=None, font_path=None, random_state=None):
+        r, g, b = colorsys.hls_to_rgb(h, l, random_state.uniform(0.2, 1))
+        return 'rgb({:d}, {:d}, {:d})'.format(int(r * 255), int(g * 255), int(b * 255))
+    return single_color_func
+
+def main():
+    prog = 'python wordcloud_cli.py'
+    description = ('A simple command line interface for wordcloud module.')
+    parser = argparse.ArgumentParser(description=description)
+    parser.add_argument('--words', metavar='file', type=argparse.FileType(),
```
pendragon-: I wonder whether this should better be called `text` instead.

pendragon-:
Apart from the naming of arguments, I think so, yes.
I did not want to expose all available options as command line arguments as that seemed to be a bit overwhelming. If you disagree I'd be open to adding those though.

wordcloud/wordcloud_cli.py
amueller: yes, probably

wordcloud/wordcloud_cli.py
```diff
@@ -0,0 +1,104 @@
...
+        color_func=color_func, background_color=args.background).generate(text)
+    image = wordcloud.to_image()
+
+    b = io.BytesIO()
+    image.save(b, format='png')
+    with imagefile:
+        imagefile.write(b.getvalue())
+
+if __name__ == '__main__':
+    main()
```
amueller: I think you should add ``args`` as a parameter for ``main``, which would allow testing.

wordcloud/wordcloud_cli.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    old_r, old_g, old_b = ImageColor.getrgb(color)
+    h, l, s = colorsys.rgb_to_hls(old_r/255., old_g/255., old_b/255.)
+    def single_color_func(word=None, font_size=None, position=None,
+                          orientation=None, font_path=None, random_state=None):
+        r, g, b = colorsys.hls_to_rgb(h, l, random_state.uniform(0.2, 1))
+        return 'rgb({:d}, {:d}, {:d})'.format(int(r * 255), int(g * 255), int(b * 255))
+    return single_color_func
+
+def main():
```
amueller: I think it would be great to add a test checking that the results of using the command line interface and the programming interface are the same. That doesn't need to be exhaustive, maybe just one or two cases.

wordcloud/wordcloud_cli.py
```diff
@@ -0,0 +1,104 @@
...
+import argparse
+import wordcloud as wc
+import numpy as np
+import sys
+import io
+import colorsys
+from PIL import Image
+from PIL import ImageColor
+
+def get_single_color_func(color):
```
amueller: This should not live here, I think.

pendragon-: Something like ImageColorGenerator from color_from_image? It's giving different saturations for a fixed hue/luminance - SaturationColorGenerator in fixed_color.

wordcloud/wordcloud_cli.py
pendragon-: What exactly do you want tested? The mapping between (commandline) arguments and the constructor arguments?
Would a mock for WordCloud be acceptable?

wordcloud/wordcloud_cli.py
amueller: What I meant was that it should not be in the CLI file. I guess all the color funcs should be in one file? You can put this in the main file for the moment if you like. But the cli file is not the right place.

pendragon-:
I've reorganized the code and added tests. Just need to understand the setup stuff & travis now.

amueller:
sweet, thanks, this is great!

pendragon-:
I've found an issue in py3 compatibility:
There's a problem with writing the content out as the output file will not accept the bytes which it tries to write. I'll provide a fix. Sorry for the inconvenience.

Issue #116, opened 2016-01-23
scale parameter may not be float


pendragon-:
If `scale` is set to a float value, it will break in `WordCloud.to_image()` as the the image size and positions calculated there need to be ints.
The question is whether to require the parameter to be `int` as it never could have been `float` before and fix the documentation, or to adapt the code to be able to handle floats.

amueller:
the code should be fixed.

amueller:
thanks :)

PR #117, opened 2016-01-26
fix handling of scale parameter & python 3 compatibility for cli

pendragon-:
- Simple solution for #116 which should work without overlap as `int` truncates.
- Adding a check for the python version when writing the output from cli.

### DIFF
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -1,5 +1,4 @@
 import argparse
-import inspect
 from wordcloud import wordcloud_cli as cli
 import wordcloud as wc
 from collections import namedtuple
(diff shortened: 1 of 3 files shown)


test/test_wordcloud_cli.py
```diff
@@ -31,22 +30,12 @@ def all_arguments():
     return arguments
 
 
-def test_argument_spec_matches_to_constructor_args():
```
pendragon-: This is in fact no longer needed as the MagicMock with `spec` ensures only arguments which actually exists are passed.

wordcloud/wordcloud_cli.py
```diff
@@ -20,10 +19,9 @@ def main(args):
         color_func=args.color_func, background_color=args.background_color).generate(args.text)
     image = wordcloud.to_image()
 
-    b = io.BytesIO()
-    image.save(b, format='png')
     with args.imagefile:
-        args.imagefile.write(b.getvalue())
+        out = args.imagefile if sys.version < '3' else args.imagefile.buffer
```
amueller: can you add a unit test for this? It wasn't tested before, right?

pendragon-: It was/still is (see `args.imagefile = NamedTemporaryFile()`) in the diff above. Unfortunately it seems that `NamedTemporaryFile` behaves differently from actual files.
While I could test to see that this part is called just the way it is now that would not ensure that the functionality works. I don't see how that would help.

amueller: Ok, I think you understand the issue better than I do ;)

pendragon-: So the tests work. I dislike how fragile the test is though - it depends on the call staying the same and inserting the Image object to write stuff out.
It could be replaced with a mock, too - I don't think the situation gets better through that though. Basically everything is reimplemented with mocks at that point, which seems wrong.
A further issue is that only a positive case is tested, while it would be nice to see it failing, too. Oh well.

pendragon-:
I've added a test where instead of mucking around with mocks, an actual file is written. That should take care of all problems.

amueller:
> That should take care of all problems.

That's very optimistic ;)

test/test_wordcloud_cli.py
```diff
@@ -106,3 +94,16 @@ def test_parse_args_defaults_to_random_color():
 
     args = cli.parse_args(['--text', text.name])
     assert_equal(args.color_func, wc.random_color_func)
+
+
+def test_cli_writes_image():
```
amueller: sweet!

PR #118, opened 2016-01-31
Feature/palette from image

nicolajkirchhof:
Hi Andreas, 

I don't know how you handle new Ideas from other users so I've just implemented the functionality that I've added for some private word cloud based projects. In essence it gives you an easy way to colorize word clouds independently to the output format. In addition to the use of a target image and its colors in defined regions, my feature gives you an easy way to color word clouds independently. 
I found http://www.colourlovers.com/palettes/most-loved/all-time/meta to be an excellent source for beautiful color palettes that produce nicely colored word clouds and using a snipping makes it very easy to get the colors and use it in your visualization.

I've tested it briefly with Python 3 (Win 10, x64) so please try it out in your environment and if you've any hints, tipps or feedback. Please let me know. 

All the best,
Nico


setup.py
```diff
@@ -3,7 +3,7 @@
 
 setup(
     name='wordcloud',
-    version='1.1.3',
```
amueller: why did you change the version?

amueller:
Thanks Nico.
This looks interesting.
Have you also checked out this example? http://amueller.github.io/word_cloud/auto_examples/colored.html
I understand yours does something different, though.
It seems a bit odd to me to use the most frequent colors, instead of using, say, kmeans.
Have you checked out this site? http://design-seeds.com/search

Creating a palette from an image seems interesting, I'm not sure your approach is the right one, though. Can you give some examples of real images and the colors you get from them, and maybe compare with k-means?

setup.py
nicolajkirchhof: Haha, this was just a habit. Sorry, I always increase the version before I add a new feature.
Shall I just reset it?

amueller: please. With that, scikit-learn would be at 0.4000 or something.

nicolajkirchhof:
Hi Andreas,

I do get your opinion but I don't think kmeans is the right way to go here. I mean you usually don't "see" the number of colors from an arbitrary image that you should know as kmeans input in order to get a good result. I've tried the site you pointed out and added an example how you could add that palettes as well. Even with the white between the colors you're still a factor of six from the six palette colors to the background. 

For real image I'd use a different approach but I'm quite skeptic that there is a good way. If you've a lot of color variations the kmean clusters will have a large spread and the colors you get out might not be what you want. Nevertheless, I had to give it a try. At https://github.com/nicolajkirchhof/PaletteFromImage  you'll find a small kmeans palette extraction example. 

In contrast to kmeans I'd suggest using a hierarchical clustering that works with a color diff. This would prevent such extreme color changes. Sure, it might create a lot of colors initially but depending on the cluster size you could rank them and select only the most common ones. Maybe I'll give that a try the next days for now I found my simple solution rather appealing. ;)

All the best,
Nico 

PS: I don't think there is a "right" approach. With clearly defined color palette images you'll be fine with or without clustering and with custom images you'll have do some manual calibration, I think.

nicolajkirchhof:
PPS: All the work is already done https://github.com/lokesh/color-thief and https://github.com/admire93/mmcq.py did it. :+1:

amueller:
cool :)
(I would have suggested going to Lap space before clustering btw).
I'm not sure I see the benefit of using "color palette images". Where do you get these images? And isn't it really easy to extract them anyhow? Maybe this shouldn't be in the wordcloud package.

nicolajkirchhof:
It's your package and therefore your decision. :)  
Since I found the provided functionality quite useful, I'll extract it and provide it in a separate module.
Please let me know if you plan to a functionality like this in any other way...

amueller:
by "this functionality" you mean extracting color maps from images that have palettes like the images you attached? This seems like a  really special case to me. It would be nice to have more / better color maps in wordcloud, but I feel this is not the way to go.
How will your package be different from mmcq?

nicolajkirchhof:
Thanks for your reply, I was in a hurry this morning and by re-reading my answer I found it not very thoughtful. 

No, I don't want to implement something that is already there. So after knowing that there is mmcq, I think I've to accept that my idea was just a quick shot and there are already better possibilities to extract palettes from unspecific images. Nevertheless, extracting them and adding them to wordclouds are two separate steps that I'd like to simplify the process as much as possible. Thus, using mmcq for color extraction and providing means to plot the extracted paletts as well as giving the user the opportunity to use a predefined callback for wordcloud would be the goal. From my POV this would enable a generic way to use "almost anything" as a color map for wordclouds. 

Apart from that, I think the second step would be to implement a good way to add predefined color palettes. Here, https://jiffyclub.github.io/palettable/ is the way to go. It has a large number of palettes, not dependencies and therefore, does not seem to be hard to include. An extended functionality might also include https://github.com/dsc/colorbrewer-python, ... 

So, for me it would be good to know what you're planning in terms of color map support within the wordcloud package. Do you, think about including a custom callback/setter for color maps (e.g. for a list of rgb triples)? Is your vision to provide some of the palletable functionality from inside wordcloud, so that a user can coose a palette by name? 

In R for instance colorbrewer lives separately from 'wordcloud' but latter provides support for adding created paletts...

amueller:
What kind of callbacks are you talking about? There is already a very generic callback, the ``color_func`` that can easily implement all possible colormaps, right?

amueller:
closing as no activity and I think this can be easily implemented with a custom ``color_func``.

PR #122, opened 2016-02-03
Added better word tokenization, though includes nltk as requirement

abhigenie92:


amueller:
I think it would be great if it was possible to pass a custom tokenizer or a custom regexp to the ``WordCloud`` class at construction. That would be more flexible than using nltk directly, and would not require nltk. Do you want to give that a shot?

amueller:
closing as I don't want to add nltk as dependency. You can already use ``from_frequencies`` or pass a custom regexp. Btw, I'd be curious in which cases this is "better" than the current tokenizer.

PR #126, opened 2016-02-11
This adds emoji support to the word cloud generator

sontek:


setup.py
```diff
@@ -11,5 +10,6 @@
                            ["wordcloud/query_integral_image.c"])],
     scripts=['wordcloud/wordcloud_cli.py'],
     packages=['wordcloud'],
-    package_data={'wordcloud': ['stopwords', 'DroidSansMono.ttf']}
+    package_data={'wordcloud': ['stopwords', 'DroidSansMono.ttf']},
+    setup_requires=['fonttools']
```
amueller: Why?

continuous_integration/install.sh
```diff
@@ -40,7 +40,7 @@ elif [[ "$DISTRIB" == "ubuntu" ]]; then
     pip install nose mock
 fi
 
-pip install -r requirements.txt
+pip install -e . -r requirements.txt
```
amueller: why that? We do the build in line 47.

wordcloud/wordcloud.py
```diff
@@ -381,16 +403,41 @@ def process_text(self, text):
...
-        flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
-                 else 0)
-        for word in re.findall(r"\w[\w']+", text, flags=flags):
+        flags = 0
+        PY3 = sys.version_info[0] == 3
+
+        if not PY3 and type(text) is unicode or PY3:
+            flags = re.UNICODE
+
+        regex = re.compile(
```
amueller: this is pretty unclear. Can you explain?

setup.py
sontek: I needed fonttools package to detect which characters are supported in the font being used so we don't try to include characters that aren't supported.  For example, the font you ship with by default has no emojis in it but `DejaVu` has a lot

wordcloud/wordcloud.py
```diff
@@ -203,6 +208,23 @@ def __init__(self, font_path=None, width=400, height=200, margin=2,
...
             font_path = FONT_PATH
         self.font_path = font_path
+        ttf = TTFont(
+            font_path, 0, verbose=0, allowVID=0,
+            ignoreDecompileErrors=True,
+            fontNumber=-1
+        )
+        supported_chars = []
+
+        for table in ttf["cmap"].tables:
```
amueller: you check whether it's in the font?

wordcloud/wordcloud.py
sontek: This is detecting the unicode range of emoji characters

amueller:
Please add documentation and test. Travis fails.
Also, I'd rather not add another dependency.

wordcloud/wordcloud.py
sontek: Yeah, if someone says :poop: but the font doesn't have it, we skip it so we don't put `[]` in the word cloud

amueller: I wouldn't put this in the init, but rather in the building of the word cloud. Also, it would be great to make this somehow optional so as not to get too many dependencies. It would also be great to factorize this functionality somewhat. Init and process text are already too long.

sontek: I put it in init so you don't have the cost of scanning the font every time you want to process the text

sontek:
@amueller I addressed most of your concerns.  I'm not sure how to make it pass on py27.  I haven't used 2.7 in a few years so I'm rusty on the differences

sontek:
@amueller This is ready for your review now I believe.

sontek:
emoji cloud

amueller:
that could looks pretty sweet, I have to say :)

setup.py
```diff
@@ -11,5 +11,7 @@
                            ["wordcloud/query_integral_image.c"])],
     scripts=['wordcloud/wordcloud_cli.py'],
     packages=['wordcloud'],
-    package_data={'wordcloud': ['stopwords', 'DroidSansMono.ttf']}
+    package_data={
+        'wordcloud': ['stopwords', 'DroidSansMono.ttf', 'Symbola.ttf']
```
amueller: have you checked the license?

amueller:
I''m a bit busy right now but I'll try to review soon. It's quite a bit of added complexity, though :-/
it would be nice if the changes could be more contained. I'm confused by conda not having fonttools. That's not a good sign, though :-/

sontek:
I don't have time to work on this anymore and it never got reviewed last year when I submitted it.  

Going to close it

PR #140, opened 2016-03-18
Add whitelist

rgaiacs:
Ensure that some word will be count.

## User case

Word cloud of programming languages that include C and R.


amueller:
sorry for the slow reply. You are changing the default behavior to using an empty stop-word list, right? Also, I would rather not add another parameter that can be implemented with one line of python (the append).

amueller:
You can just use your own tokenizer by using ``from_frequencies`` btw.

PR #160, opened 2016-07-16
Added more words to list of stopwords

piyushkhemka:
Added more words to the list of stopwords.

I think the original list was the 'Default English stopwords list' from [Ranks.nl](http://www.ranks.nl/stopwords). I have replaced it with 'Long Stopwords List'


amueller:
I'm not sure this is a good default. It has "information" and "proud" among others. What makes these stopwords?

piyushkhemka:
You are correct. Words like proud & information shouldn't be in the list of stopwords. Sorry I didn't notice them in the updated list.

I was looking to make the current list more comprehensive because it misses out on a lot of stopwords like - shall, aren't, k, may etc. (In one of the word clouds I made, shall came up as the biggest sized word in the whole cloud)

I will try to find a better list of stopwords on the internet and modify the pull request.

amueller:
thanks :)

piyushkhemka:
Hey Andreas

I couldn't find a better stopwords list online. The lists I found contained a lot of verbs, nouns which you didn't want in the stopwords list.

So, I manually added a few words which were coming up in the word clouds I made like shall for example.

I also sorted the list (using some online tool) alphabetically which is why you might see a few words jumbled up and down.

amueller:
looks good, thanks :)

PR #182, opened 2016-10-31
Add a colored by group example

vkolmakov:
I was tasked with making a word cloud that had certain groups of words colored with certain shades and figured that somebody else might find this example useful.


amueller:
nice example. Maybe it would be cleaner as a class with a ``__call__`` though:
```python
class grouped_color_func(object):
    def __init__(self, word_to_color, default="white"):
        self.word_to_color = word_to_color
        self.default = default
    def __call__(self, word, **kwargs):
        if word in self.word_to_color:
            return self.word_to_color[word]
        return self.default
````

**Comment by vkolmakov** [comment]
@amueller Thanks for the feedback, __call__ does fit in very well here.

However, the main thing I was trying to display in this example is how to use a single color function (from get_single_color_func) to color a group of words with various shades of the supplied color.

I can make two classes, name them along the lines of GroupedColorFuncSimple (which will use the exact colors and build a dictionary from word to color inside of it) and GroupedColorFunc (which will use a single_color_func to color the group of words)

**Comment by amueller** [comment]
Sorry, I'm not quite sure I understand what you're trying to show. Why do you need ``get_single_color_func``?

**Comment by vkolmakov** [comment]
I'm trying to create one `single_color_func` for each group of words and use it on them. That produces a word cloud with selected words being colored with various shades (same hue/saturation + different values in hsv) of the specified color. This might be useful if there is a small number of groups and a lot of words, although could be harmful for smaller examples.

Let me know if you would want me to remove it and just keep the simple example.

**Comment by amueller** [comment]
Ah, sorry, I didn't get your original example entirely. I mean the example is fine but I'm not sure if this is something very widely used... Do you have a use-case example?
I would maybe rather see a topic model trained on the paragraphs of alice in wonderland and then color the words by topics?

**Comment by vkolmakov** [comment]
It does work very well if there are a lot of words, small range for word frequencies and a small number or groups. Without the use of `single_color_func` parts of the cloud would usually turn into single color blobs and it's hard to distinguish words.

I do see your point though, this is indeed a very specific use-case, and it would definitely make a lot more sense to keep the examples focused. I can remove the class that uses `single_color_func` and just keep the one that colors specific groups of words with exact colors. Let me know what do you think about it.

The idea of coloring words by topics sounds awesome, although it's a bit out of reach for me.

**Comment by amueller** [comment]
How do you assign the words to the groups if there are a lot of words? I might just write an example with the topics. If you do the single color func, I'm more inclined to merge this because it illustrates how to create a custom color function that is a bit more advanced.

**Comment by vkolmakov** [comment]
In my case I was provided with words to be colored in separate files (e.g. `red.txt`, `green.txt`, etc), which is basically the same thing I have here: color provided to `single_color_func` is mapped to a list of words.

I guess that a more general example with use of `single_color_func` will include mapping colors to predicates, and color words that match the predicate function, something along the lines of:

```python
is_name = lambda w: w in ['John', 'Jane', ...]
starts_with_vowel = lambda w: w[0].lower() in ['a', 'e', 'o', 'i', 'o', 'u']
starts_with_consonant = lambda w: not starts_with_vowel(w)

color_to_predicate = {
  'red': is_name,
  'green': starts_with_vowel,
  'blue': starts_with_consonant
}

# pass color_to_predicate to the constructor
# created function will color names with red single_color_func,
# words that start with vowels with green, and words that start
# with consonants with blue
```

This way it will be possible to cover pretty much any conditional coloring scheme. Although general, it might be a bit more confusing than simple coloring by membership using a `single_color_func`. The latter is already implemented in the last commit, let me know what do you think about it, and if you think that coloring by predicates will be more suitable.

amueller:
I guess the example is alright as it is.

PR #189, opened 2016-11-14
Added a part about the command-line tool in README

### LINKED ISSUE #187: Mention the command-line interface
Hello!

I've just used word_cloud on the command-line and it works very well. However, the README and the site don't mention this command-line tool and I had to look at the source directory to see that there was a `wordcloud_cli.py` file and that it was installed with the library.

Is there a reason this tool is not mentioned anywhere (like is it unstable or something?) If that's not the case then it would be a nice idea to write a few words about it in the README, so that people know it exists.

foobuzz:
This fixes #187


amueller:
sweet, thanks :)

PR #190, opened 2016-11-19
Add version info to module and cli

michaelaye:
* added the standard __version__ string to __init__.py
* added setup.cfg (future proof) that has the bumpversion setup
* added the version string info to the argparse section in the cli

Thanks for your nice tool!

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,12 @@
+[bumpversion]
+current_version = 1.2.1
+commit = True
+tag = True
+
+[bumpversion:file:setup.py]
+search = version='{current_version}'
+replace = version='{new_version}'
+
+[bumpversion:file:wordcloud/__init__.py]
+search = __version__ = '{current_version}'
+replace = __version__ = '{new_version}'
--- a/wordcloud/__init__.py
+++ b/wordcloud/__init__.py
@@ -2,3 +2,5 @@ from .wordcloud import WordCloud, STOPWORDS, random_color_func, get_single_color
 from .color_from_image import ImageColorGenerator
 
 __all__ = ['WordCloud', 'STOPWORDS', 'random_color_func', 'get_single_color_func', 'ImageColorGenerator']
+
+__version__ = '1.2.1'
--- a/wordcloud/wordcloud_cli.py
+++ b/wordcloud/wordcloud_cli.py
@@ -51,6 +51,8 @@ def parse_args(arguments):
         help='use given color as coloring for the image - accepts any value from PIL.ImageColor.getcolor')
     parser.add_argument('--background', metavar='color', default='black', type=str, dest='background_color',
         help='use given color as background color for the image - accepts any value from PIL.ImageColor.getcolor')
+    parser.add_argument('--version', action='version',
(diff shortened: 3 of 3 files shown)


amueller:
Thanks for the PR. Travis fails ;)

What is your motivation for using setup.cfg? That seems like another level of indirection. Can we not import it from ``__init__.py`` into ``setup.py``? Or does that fail because of the imports from the build package? Then we could add a ``_version.py`` or something that just defines the version and import that both in ``__init__.py`` and ``setup.py``?

michaelaye:
It's a matter of taste I guess. One shouldn't import your own module in setup.py, creates lots of havoc for people installing it fresh.
Which means you have to write a bit of file reading code in setup.py to read out the version string from version.py. I'd rather have bumpversion manage two files than have extra code in setup.py to parse version.py but if you prefer that I can change it. 
I'm not sure I understand your question concerning setup.cfg. Are you asking why I use that instead of .bumpversion.cfg or why I use bumpversion at all? If it's the former, setup.cfg is the way to go these days it seems, as most untold now can read their confit out of that file. Much nicer to have one config file than one for each utility.

amueller:
I guess why use bumpversion ;) I have not seen it before. Do you consider it widely used?

amueller:
And I'm fine with keeping it as-is (apart from the travis error)

amueller:
> If it's the former, setup.cfg is the way to go these days it seems, as most untold now can read their confit out of that file. Much nicer to have one config file than one for each utility.

Sorry, I can't parse that.

I just feel bumpversion seems pretty obscure. The build is failing and it has less github stars than this package.

michaelaye:
Wow, that's some serious autocorrect fail. ;)

What I wanted to write is:
> If it's the former, setup.cfg is the way to go these days it seems, as most utils now can read their config out of that file. Much nicer to have one config file than one for each utility.


Bumpversion is easier to use, simpler idea, than versioneer.py (assuming that you are fine with leaving the world of manually managing version strings? ;).

The build only fails for Python2, which smells like I can't do some import like I do in Python2, not something related to how I manage the version string. It's long since I developed on Python2, am Py3 only for years. I'll see what I can find out to fix this.

amueller:
I actually think most packages do version strings manually. Changing the version in multiple places is error-prone though. I have not really been in touch with any version string managing tools, and it looks like there is no standard (versioneer also doesn't seem like a very popular project).
The problem with using any of these tools is that it adds another layer of abstraction with makes it harder to maintain the package and increases the barrier of entry for new developers.

It looks like versioneer actually adds a ``_version.py`` file, like I suggested, only that it contains some magic related to the version control.

Can you give examples of major packages using bumpversion or versioneer?

bumpversion actually looks nice, so after reading the docs I'm less skeptical ;)
It basically does a search and replace and additional tagging, right?
So it does
```
git ls-files | xargs sed -i -e 's/OLDVERSION/NEWVERSION/g'
git tag NEWVERSION
```
and because the oldversion is read from the configfile, it will also be replaced there, which is nice...
So I'm fine with using it, but I'm still curious about whether it's used elsewhere.

amueller:
I prefer #297

PR #198, opened 2016-12-01
Fix install

remram44:
* Unify your requirements
* Add version constraint for numpy for pip and conda
* Fix conda recipe (did this build before without the `--single-version-externally-managed` trick?)

I'm not sure you want matplotlib to be a hard requirement since you went to the trouble of not having it be a top-level import.


amueller:
thanks. how about we get rid of requirements.txt?

remram44:
Done!

amueller:
sweet!

amueller:
oh you need to fix the travis script now, I guess...

amueller:
thanks!

PR #209, opened 2017-01-01
make wc.width and wc.height into property

Xunius:
fetch wc.mask for width and height queries if mask exists, otherwise use user specified shapes. This is mainly for better consistency, ease of use and to help clarify ambiguities:

e.g. `wc.width` now gives same result as `wc.mask.shape[1]`, so the user can use the former. The latter form is less intuitive and not obvious unless one reads into the code.
e.g. if a `wc` obj was created using default shape, and a mask was assigned later that has a different shape, now the shape of `wc` gets updated by new shape of the mask.

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -283,8 +283,8 @@ class WordCloud(object):
         self.colormap = colormap
         self.collocations = collocations
         self.font_path = font_path
-        self.width = width
-        self.height = height
+        self._width = width
+        self._height = height
         self.margin = margin
         self.prefer_horizontal = prefer_horizontal
         self.mask = mask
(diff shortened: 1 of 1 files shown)


wordcloud/wordcloud.py
```diff
@@ -310,6 +310,20 @@ def __init__(self, font_path=None, width=400, height=200, margin=2,
                           " it had no effect. Look into relative_scaling.",
                           DeprecationWarning)
         self.normalize_plurals = normalize_plurals
+        
+    @property
```
amueller: why not just set it in the constructor based on ``mask``? In case someone changes ``mask`` later? Can you please also add a test?

Btw, it might also be helpful to set ``height`` and ``width`` to ``None`` and if no mask is provided set them to the default value, and if a value is provided and a mask is also provided raise an error. that should make debugging easier maybe?

Issue #202, opened 2016-12-02
Why a WordCloud shouldn't render numbers?


leferrad:
In some cases, I need to compute a WordCloud from text with numbers (e.g. dates, numerical categories, etc), but I couldn't do that because the following snippet on the code:

https://github.com/amueller/word_cloud/blob/master/wordcloud/wordcloud.py#L514-L515

Is there some reason for removing number in the word_cloud? In my own version, I've just commented that, but I wonder the reason of this design decision in WordCloud.

Thanks!

amueller:
usually they are not interesting. Feel free to add a parameter to make that behavior optional.

jbbqqf:
Hi! Triaging older issues — the toggle @amueller suggested here ("Feel free to add a parameter to make that behavior optional") has been added.

Evidence:
- `WordCloud` now accepts `include_numbers=False` as a constructor kwarg ([wordcloud/wordcloud.py#L318](https://github.com/amueller/word_cloud/blob/93c5e4cd50c5f2c56ec7eb624a620790dab5df6c/wordcloud/wordcloud.py#L318)), documented in the class docstring as *"include_numbers : bool, default=False — Whether to include numbers as phrases or not."* ([L272-L273](https://github.com/amueller/word_cloud/blob/93c5e4cd50c5f2c56ec7eb624a620790dab5df6c/wordcloud/wordcloud.py#L272-L273)).
- The numeric filter in `process_text` is now gated on that flag: `if not self.include_numbers: words = [word for word in words if not word.isdigit()]` ([L591-L592](https://github.com/amueller/word_cloud/blob/93c5e4cd50c5f2c56ec7eb624a620790dab5df6c/wordcloud/wordcloud.py#L591-L592)). Setting `include_numbers=True` lets numbers through.
- The CLI exposes the same toggle as `--include_numbers` ([wordcloud/wordcloud_cli.py#L159-L163](https://github.com/amueller/word_cloud/blob/93c5e4cd50c5f2c56ec7eb624a620790dab5df6c/wordcloud/wordcloud_cli.py#L159-L163)).

If there was a broader follow-up tracked here (e.g. a `min_word_length`-style threshold for numeric tokens) that I'm missing, happy to dig further. Otherwise would you mind closing this out?

---
*Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually against current `main`.*

PR #215, opened 2017-01-07
Adding support to render numbers

leferrad:
According to the issue #202, this refactor only adds a boolean argument 'render_numbers' in WordCloud class for indicating if the words which are numbers should be deleted or not.

2 core affected lines:
https://github.com/amueller/word_cloud/blob/master/wordcloud/wordcloud.py#L272 (WordCloud.__init__)
https://github.com/amueller/word_cloud/blob/master/wordcloud/wordcloud.py#L514-L515 (WordCloud.process_text)

Regards

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -241,6 +241,10 @@ class WordCloud(object):
         is removed and its counts are added to the version without
         trailing 's' -- unless the word ends with 'ss'.
 
+    render_numbers : bool, default=False
+        Whether to include numbers as words for the WordCloud. This could
+        be useful in case of handle categorical data or dates.
+
     Attributes
     ----------
     ``words_`` : dict of string to float
@@ -269,7 +273,7 @@ class WordCloud(object):
                  stopwords=None, random_state=None, background_color='black',
                  max_font_size=None, font_step=1, mode="RGB",
                  relative_scaling=.5, regexp=None, collocations=True,
-                 colormap=None, normalize_plurals=True):
+                 colormap=None, normalize_plurals=True, render_numbers=False):
         if font_path is None:
             font_path = FONT_PATH
         if color_func is None and colormap is None:
@@ -310,6 +314,7 @@ class WordCloud(object):
                           " it had no effect. Look into relative_scaling.",
                           DeprecationWarning)
         self.normalize_plurals = normalize_plurals
+        self.render_numbers = render_numbers
 
     def fit_words(self, frequencies):
         """Create a word_cloud from words and frequencies.
@@ -511,8 +516,9 @@ class WordCloud(object):
         # remove 's
         words = [word[:-2] if word.lower().endswith("'s") else word
                  for word in words]
-        # remove numbers
-        words = [word for word in words if not word.isdigit()]
+        if self.render_numbers is False:
+            # remove numbers
+            words = [word for word in words if not word.isdigit()]
 
         if self.collocations:
             word_counts = unigrams_and_bigrams(words, self.normalize_plurals)
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -508,7 +508,7 @@ class WordCloud(object):
 
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
                  else 0)
-        regexp = self.regexp if self.regexp is not None else r"\w[\w']+"
+        regexp = self.regexp if self.regexp is not None else r"(\w[\w']+|\w+)"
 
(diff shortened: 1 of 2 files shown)


amueller:
Sorry for the late reply. Looks good, but could you please add a test?

leferrad:
Yes of course, sorry. Now there is a new test for that, and I also changed the regexp used in WordCloud.process_text for accepting numbers too. Please review that and tell me if it's right.

wordcloud/wordcloud.py
```diff
@@ -503,16 +508,17 @@ def process_text(self, text):
 
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
                  else 0)
-        regexp = self.regexp if self.regexp is not None else r"\w[\w']+"
+        regexp = self.regexp if self.regexp is not None else r"(\w[\w']+|\w+)"
```
amueller: Can you explain your change? I think I wouldn't change the current default behavior. Is the intent that you want single digits to show? We could change it when  render_numbers is false but that would create dependencies between the parameters, which I don't like.

Maybe the simplest solution is to undo this change and just put this as a custom regexp in the test? Or use longer numbers in the test?
Thanks!

leferrad: Yes, I didn't want to change the current default behavior. But using the current regex I couldn't take any number because it filters them. 

Therefore I've just add a kind of filter in the regex: if the token doesn't match using the first part (the current `\w[\w']`), it tries with the second one (`\w+`). You can test this regex with samples of text if you're not convinced, for instance in this link: https://regex101.com/. But you can see that it's just a modification to let numbers be matched (and they will be removed in the snippet https://github.com/leferrad/word_cloud/blob/number_rendering_support/wordcloud/wordcloud.py#L519-L521 if the argument `render_numbers=False`), but it doesn't cancel the previous regex behavior I guess. 

I'm agree with the fact that it's not fine to "switch" a regex based on the parameter `render_numbers` because it could fork the behavior, so that's the reason why I've just override the current regex. However, if you still uncomfortable with that override, and I can look for another way to achieve this intended feature.

amueller: Well it does overwrite the current behavior - you could say it does so by design. The current regex doesn't allow 1 character words, and you changed it so that it does allow one-character words. That makes the first part of the regex pretty superfluous, doesn't it?

leferrad: All right, I didn't notice that you really want to filter out that words. Can you give me other use cases for the matching desired? So far, that case of 1 char words can be avoided with a filter using the condition `len(word)==1 and not word.isdigit()`, but if you're thinking in more cases then it's not worth to "hard code" the filtering process.

amueller: if you add another "if" that requires another option, so I'd rather avoid it. The regex is very flexible and I'd like to include much of the options there.
We could actually try changing the regexp for digits and not use the ``isdigit`` function maybe?

leferrad: So you propose to make an additional regex for digits in case of having `render_numbers=True` or what is the idea? Because  the regex that I proposed accept numbers (`\w+` matches any word character, and it's equal to `[a-zA-Z0-9_]`) but it also needs an additional step to remove 1 char words. Please explain me more about your thoughts in order to know which should the best way to achieve this `render_numbers` feature.

amueller: The idea was not to add a ``render_numbers`` parameter but include numbers directly via the regexp, like this: http://stackoverflow.com/questions/3831051/regex-how-to-match-a-string-that-is-not-only-numbers

leferrad: Sorry for be late, I was completing my career studies and then on vacations. Let's make it clear, and list which cases should cover the regex in order to know how to build it (you should say which ones are desirable):

**ACCEPT**
- [ ] words with more than 1 char, possibly containing numbers (e.g. "word", "w0rd").
- [ ] words having the character ' literally (e.g. "O'Reilly", "didn't")
- [ ] numbers (e.g. "1", "2", "2017")

**REFUSE**
- [ ] words with a single char which is not a number (e.g. "A", "i", "x")

That makes sense? Once we know the expected behavior of the regex, then we can build it and test all the code with those changes.

amueller: I wouldn't want the current default behavior to change, so I would but the numbers in "refuse" for the default regex, but add to the documentation an alternative regex to use that allows numbers.

amueller:
@arubiales you would need this branch. Though I'm not more inclined towards using the approach implemented here. But I don't have time to look into it.

amueller:
fixed in #465

PR #223, opened 2017-01-21
Transparent wordcloud example.

pylipp:
A black-and-white wordcloud is generated from wordcloud's source code and substituted as the alpha channel of the background image. This makes the wordcloud text show the underlaying image while the non-text part of the image is white. 
An option is to show the entire background image and 'dim' the transparency of the wordcloud text.
The example is tested on Ubuntu 16.04 using Python 3.


amueller:
You could just set background to be transparent and image mode to RGBA. Also, I think this is not really worth an example, there's already too many.

Issue #226, opened 2017-01-25
generate from list of weighted terms instead of full text only


rubienr:
The **wordcloud_cli.py** could be extended to read from a weighted term list instead of full text only. It makes sense if one knows the terms and their weights beforehand. 

Example:
```
wordcloud_cli.py --imagefile foo.png --text terms.txt --csv
```
**terms.txt** would list the terms and their weights:
```
term, weight
term1, 2
term two, 1
...
term n, 3
```

**Workaround**:
```
def main(args):
    wordcloud = wc.WordCloud(stopwords=args.stopwords, mask=args.mask,
        width=args.width, height=args.height, font_path=args.font_path,
        margin=args.margin, relative_scaling=args.relative_scaling,
        color_func=args.color_func, background_color=args.background_color)
    if not args.csv:
        wordcloud = wordcloud.generate(args.text)
    else:
        wordcloud = wordcloud.generate_from_frequencies(args.text)
...
```

```

def parse_args(arguments):
...
    parser.add_argument('--csv', default=False, action='store_true',
        help='indicates text file contains CSV of weighted terms; 1st column: terms, 2nd colum: weight as integer, 1st row: column names')
...
    if not args.csv:
        with args.text:
            args.text = args.text.read()
    else:
        reader = csv.DictReader(args.text)
        args.text = list()
        for row in reader:
            args.text.append((row[reader.fieldnames[0]], int(row[reader.fieldnames[1]])))
...
```

I would classify the issue's priority as low. In case one finds this feature request irrElephant, simply close this issue.

tnx,
r²

amueller:
I think it's a good idea. PR welcome :)

zouzias:
Hi all,

Is anyone working on this issue? I would happy to give it a try.

amueller:
There is some work in #230 but that seems stalled.
I'm not sure if I would use the argument names given in the description. Maybe instead of ``--text terms.txt --csv`` I'd do ``--weights terms.txt`` or ``--counts terms.txt``

victoriastuart:
Agreed: this would be  useful feature!

PR #230, opened 2017-02-15
Added option to use files with (word , weight) as described in Issue #226

mohammadKhalifa:


wordcloud/wordcloud.py
```diff
@@ -538,7 +540,13 @@ def generate_from_text(self, text):
         -------
         self
         """
-        words = self.process_text(text)
```
amueller: you don't need to change the wordcloud class. You can just use ``generate_from_frequencies``.

mohammadKhalifa:
I know. But I didn't want to expose it in the wordcloud_cli.py. Do you think its okay?

wordcloud/wordcloud.py
mohammadKhalifa: I know. But I didn't want to expose it in wordcloud_cli.py. Do you think its okay?

mohammadKhalifa: I know. But I didn't want to expose it in wordcloud_cli.py. Do you think its okay?

amueller: it's not internal, it's public and documented as such.

mohammadKhalifa:
Okay. Will I need to add `ArgOption` in `test_wordcloud_cli.py` for my new argument  ?

amueller:
Sorry I'm not sure I understand the question.

mohammadKhalifa:
I have the following error in one of the build checks 
```
ERROR: test_wordcloud_cli.test_main_passes_arguments_through
----------------------------------------------------------------------
  File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/amueller/word_cloud/test/test_wordcloud_cli.py", line 45, in test_main_passes_arguments_through
    cli.main(args)
  File "/home/travis/build/amueller/word_cloud/wordcloud/wordcloud_cli.py", line 20, in main
    weightedwords = args.weightedwords).generate(args.text)
AttributeError: 'Namespace' object has no attribute 'weightedwords'
```

AS far as I know, I need to add `weightedwords` to the arguments in `test_wordcloud_cli.py`

mohammadKhalifa:
The problem with using `generate_from_frequencies` right away is that a user of the `WordClass` method will have to create the weights dict and pass it to the method themselves which would restrict the feature to users of the cli only. I believe I have to modify `WordCloud` to allow user to use this feature immediately by setting the `weightedwords` arg and passing the text to `generate`. Based on the argument `generate` creates the dict and passes it to `generate_from_frequencies`

amueller:
Sorry I don't get your point. This feature is already available for users of the WordCloud class as ``generate_from_frequency``.

setup.py
```diff
@@ -9,10 +9,11 @@
     url='https://github.com/amueller/word_cloud',
     description='A little word cloud generator',
     license='MIT',
-    install_requires=['matplotlib', 'numpy>=1.6.1', 'pillow'],
+    install_requires=['matplotlib', 'numpy>=1.6.1', 'pillow', 'python-bidi', 'arabic-reshaper'],
```
amueller: These shouldn't be general dependencies.

amueller:
this branch now seems to contain something else entirely?!

mohammadKhalifa:
Yes I've been woking on Arabic language support and I commited to this PR by mistake. I will reset this branch.

mohammadKhalifa:
@amueller  What I'm trying to say is that `generate_from_frequencies` requires a `dict`. I want to spare the user to have to create the dict so they could just pass the csv text to the `WordCloud` object and setting the `weightedword` argument to be true.

amueller:
I think if people use the programmatic interface they can do the conversion themselves. Adding all possible interfaces to the library complicates the code with near zero benefit to the user.

PR #234, opened 2017-03-03
Add support for emoji.

rmelick:
The example emoji.py works on Ubuntu with the Symbola font.  It seems
there may be some bugs in PIL / Pillow that prevent the OS X emoji
fonts from functioning correctly.


wordcloud/wordcloud.py
```diff
@@ -507,9 +508,20 @@ def process_text(self, text):
 
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode
                  else 0)
-        regexp = self.regexp if self.regexp is not None else r"\w[\w']+"
 
+        # the regex used to detect words is a combination of normal words, ascii art, and emojis
```
amueller: I would rather not include such a complex regexp. You can just pass it as parameter for the example, right?

rmelick: Great idea.   I've updated to just pass the fancy regex as a parameter.

amueller:
thanks!

PR #242, opened 2017-03-21
Move check for empty frequencies before first access, and raise ValueError if check fails.

### LINKED ISSUE #201: Checking 'frequencies' list in function WordCloud.generate_from_frequencies()
I'm having an issue related to the generation of the word_cloud from a list of frequencies per word. As you can see below, the function is trying to retrieve the max frequency from a list 'frecuencies', but it doesn't check before if there is some element in that list:

https://github.com/amueller/word_cloud/blob/master/wordcloud/wordcloud.py#L350

That kind of check is done after, in the following snippet (but it's too late, because we already tried to access to the first element on a probably empty list):

https://github.com/amueller/word_cloud/blob/master/wordcloud/wordcloud.py#L360-L362

In my own version, I've done the following correction related to the first snippet in order to avoid some crashes (just as an example):

``` 
        try:
            max_frequency = float(frequencies[0][1])
        except:
            print "ERROR with list 'frecuencies'=", frequencies
```

Best regards!

cjmay:
Fixes #201 by moving check just after `frequencies` definition and raising a `ValueError` if the check fails (rather than just printing).  I've always felt a little unsure about which exception to raise, but raising an exception seems like the desired behavior here (rather than just printing an error to stdout)?  The added tests fail before the patch is applied.

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -76,6 +76,16 @@ def test_multiple_s():
     assert_in("flosss", wc.words_)
 
 
+def test_empty_text():
+    # test originally empty text raises an exception
+    wc = WordCloud(stopwords=[])
+    assert_raises(ValueError, wc.generate, '')
+
+    # test empty-after-filtering text raises an exception
+    wc = WordCloud(stopwords=['a', 'b'])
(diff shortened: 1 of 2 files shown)


amueller:
thanks :)

PR #243, opened 2017-03-21
Make collocations in word cloud optional in CLI.

### LINKED ISSUE #241: Some text distributions yield duplicate words in image
As a workaround for #226 I tried generating text according to the desired distribution: [food.txt](https://github.com/amueller/word_cloud/files/858596/food.txt)

Note in particular there are few word types but relatively high number of tokens per type compared to natural language.  Tokens of each word type are also grouped.  (My guess would be one or both of these properties are causing the issue.)

However, `wordcloud_cli.py` (from master) on this generated text file, with default options, yielded a word cloud with each word duplicated conspicuously: food

cjmay:
From my perspective making collocations off by default is more intuitive (it fits with the behavior of leaving stop word filtering off---an empty stop word list---by default).  My expectations of the program are: here are some words, make a word cloud out of them, don't do anything surprising.  But if you don't agree with this, I still think the `collocations` flag should be exposed in the CLI (perhaps the alternative would be a `--no-collocations` argument), as I need it in my use case.

Closes #241.

### DIFF
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -17,7 +17,10 @@ ARGUMENT_SPEC_TYPED = [
     ArgOption(cli_name='width', init_name='width', pass_value=13, fail_value=1.),
     ArgOption(cli_name='height', init_name='height', pass_value=15, fail_value=1.),
     ArgOption(cli_name='margin', init_name='margin', pass_value=17, fail_value=1.),
-    ArgOption(cli_name='relative_scaling', init_name='relative_scaling', pass_value=1, fail_value='c')
+    ArgOption(cli_name='relative_scaling', init_name='relative_scaling', pass_value=1, fail_value='c'),
+]
+ARGUMENT_SPEC_UNARY = [
+    ArgOption(cli_name='no_collocations', init_name='collocations', pass_value=True, fail_value=1)
 ]
 ARGUMENT_SPEC_REMAINING = [
     ArgOption(cli_name='stopwords', init_name='stopwords', pass_value=temp.name, fail_value=None),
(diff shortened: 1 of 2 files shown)


amueller:
This PR only adds the flag, but keeps the default as "true", right?
That seems like a good solution to me. The program is more "here is some text, make a word-cloud out of it". And separating two words that always appear together isn't that intuitive to me.

cjmay:
Okay, I'll change it: the default behavior in the CLI would be switched in the current PR.

amueller:
thanks :)

cjmay:
The revised PR adds a unary `--no-collocations` flag to the CLI that, when specified, passes `collocations=False` to `WordCloud`.

amueller:
Thanks :)

pybokeh:
Just making sure I understand the future direction, will collocations be set to False by default?  If so, I definitely prefer this.  I ran some old code and couldn't figure out for the life of me why I was having duplicate words showing up.  Now I know!

amueller:
@pybokeh no it will not be off by default. Why did you have duplicate words? Did you sort the words before passing them to the library? If so, why not directly provide word counts?

pybokeh:
No, no sorting.  For exame, I did alice example without adding "said" to
stopwords and got Alice and said showing up twice.  This is very unexpected
for me, but maybe not for others.

On May 1, 2017 5:30 PM, "Andreas Mueller" <notifications@github.com> wrote:

@pybokeh <https://github.com/pybokeh> no it will not be off by default. Why
did you have duplicate words? Did you sort the words before passing them to
the library? If so, why not directly provide word counts?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/amueller/word_cloud/pull/243#issuecomment-298438042>,
or mute the thread

amueller:
You mean you had "Alice", "said" and "Alice said" show up, right?

pybokeh:
Yes, sorry did not list all the combinations.

Issue #253, opened 2017-04-18
WordCloud doesn't support unicode stopwords


icyblade:
When unicode stopwords are passed into `WordCloud`, error will be raised:

    TypeError: descriptor 'lower' requires a 'str' object but received a 'unicode'

amueller:
Are you on python2 or python3? On Python3 this should work without issue.

icyblade:
I'm using python 2.7. In python 2, `unicode` is different from `str` which leads to this problem.

jb6:
I have the same problem as icyblade, how was that resolved ? My stopwords in French are not being stopped (I did transform the unicode words in a set of strings, but the words are still added to the worldcloud). Any suggestion ?

amueller:
@jb6 Can you open a new issue and provide code to reproduce? The easiest way it probably to upgrade to  Python3.

jb6:
yes sure, will do. Just to let you know that I have just tried on Python 3 and it does work nicely. The issue therefore concerns specific stopwords (for example words with accents) that are not "stopped" and appear in the cloud when running **Python 2.7** .

PR #254, opened 2017-04-18
add unicode support for stopwords

icyblade:
Trying to solve https://github.com/amueller/word_cloud/issues/253

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -246,3 +246,14 @@ def test_relative_scaling_zero():
     # non-regression test for non-integer font size
     wc = WordCloud(relative_scaling=0)
     wc.generate(THIS)
+
+
+def test_unicode_stopwords():
+    wc_unicode = WordCloud(stopwords=[u'Beautiful'])
+    words_unicode = wc_unicode.process_text(unicode(THIS))
+
+    wc_str = WordCloud(stopwords=['Beautiful'])
+    words_str = wc_str.process_text(str(THIS))
+
+    assert_true(words_unicode == words_str)
+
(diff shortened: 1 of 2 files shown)


amueller:
Maybe use
```python
stopwords = set([x.lower() for x in self.stopwords])
```
instead?
Also please add a non-regression test

amueller:
happy to merge this if you can add a test :)

icyblade:
Before PR:

    $ python -c "from wordcloud import WordCloud; print(WordCloud(stopwords=[u'test']).generate(u'this is a test of this software'))"
      File "/usr/lib64/python2.7/site-packages/wordcloud/wordcloud.py", line 556, in generate
        return self.generate_from_text(text)
      File "/usr/lib64/python2.7/site-packages/wordcloud/wordcloud.py", line 541, in generate_from_text
        words = self.process_text(text)
      File "/usr/lib64/python2.7/site-packages/wordcloud/wordcloud.py", line 506, in process_text
        stopwords = set(map(str.lower, self.stopwords))
    TypeError: descriptor 'lower' requires a 'str' object but received a 'unicode'
    $
    
After PR:

    $ python -c "from wordcloud import WordCloud; print(WordCloud(stopwords=[u'test']).generate(u'this is a test of this software'))"
    <wordcloud.wordcloud.WordCloud object at 0x7f69d930e910>
    $

amueller:
Please add that to ``test/test_wordcloud.py`` as a separate test.

amueller:
Thanks. Unfortunately this test doesn't pass on Python 3 because there is no ``unicode`` type.
We could either skip this test on python3 or try to make it compatible.

icyblade:
As I have no py3 environment, I gotta deal with this problem tomorrow.

amueller:
no worries, thanks for your help!

amueller:
Thanks :)

PR #261, opened 2017-05-09
Example of circular word cloud

jerry-shijieli:
Another example to make circular word cloud using a black-white circle picture mask


amueller:
thanks but I feel that's pretty redundant with the storm trooper example. Can you say what functionality this example shows in particular? I don't want to add too many examples.

Issue #238, opened 2017-03-19
Different Languages


nikusha37:
Hello. Thank you for this great library.

The issue is that I am trying to create it from Georgian words (ქართული სიტყვები). Goergian has its own alphabet. But it is showing me rectangular images instead of words.

Can I do it with this library?

amueller:
you just need to use a font that contains these characters.

amueller:
I think it would be good to use a default font that covers "most languages" as much as possible, see https://github.com/amueller/word_cloud/pull/219#issuecomment-272301932

amueller:
also see discussion in #315 on arabic fonts.

amueller:
@lingdoc yes I've looked into those a while ago. Unfortunately they are too big to include in the package (noto, that is). Also, they are per-language. I thought about combining several of the noto fonts into a single font to ship with wordcloud, but it would be hard to decide what to include, and Chinese is pretty big (and one of the most asked-for languages)

lingdoc:
@amueller understood - I'm definitely not suggesting you include them directly in the package. The links themselves, though, might be useful resources to include in the documentation or on the landing page for users who want to create wordclouds for languages with a non-Roman font.

amueller:
Pr to the docs very welcome! We should have a FAQ

PR #270, opened 2017-05-28
add an example of showing Chinese character

xmanrui:
Add a font to support showing Chinese character.
Add an example of showing Chinese character.

### DIFF
--- a/examples/show_Chinese.py
+++ b/examples/show_Chinese.py
@@ -1,0 +1,40 @@
+# -*- coding: utf-8 -*-
+
+from wordcloud import WordCloud
+import matplotlib.pyplot as plt
+
+text = '''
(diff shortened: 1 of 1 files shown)


amueller:
Hey. Thank you for the example. I was wondering, do you know of an open font that supports a wide set of languages so we could use that as default? Also see #219 and #238. I think NotoSans might qualify, but I didn't have time to look into it.
It would be great if most languages worked "out of the box" with the default font.

jackpgao:
It seems doesn't work.
The Chinese not show in picture.

amueller:
I think we might merge #315 instead. Arabic also allows for tokenization, for Chinese we really can't do tokenization in wordcloud from what I understand.

PR #289, opened 2017-08-24
Update README.md

sluetze:
Added python devel for pyconfig.h on CentOS / RHEL compilation error.

dunno how many people use it on ServerOS, i do :)


amueller:
thanks :) [Ideally we'd have a wheel or (better) conda packages]

PR #295, opened 2017-09-09
Add option to override regular expression to cli

### LINKED ISSUE #224: regexp cli argument for wordclout_cli.py
I faced some limitations when using wordcloud_cli.py. The default regexp  \w[\w']+ does not match i.e. "c" "c++", "R", "µC", see [1].  I would recommend to register a new optional command line argument and pass it to the WordCloud constructor.

```
parser.add_argument('--regexp', metavar='regexpression', default="\w[\w']+",
help="specify an explicit regular expression, default is \w[\w']+")
wordcloud = wc.WordCloud(..., regexp=regexpression).generate(args.text)
```
[1] http://fiddle.re/60e26n

boidolr:
fixes #224

### DIFF
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -24,6 +24,7 @@ ARGUMENT_SPEC_UNARY = [
 ]
 ARGUMENT_SPEC_REMAINING = [
     ArgOption(cli_name='stopwords', init_name='stopwords', pass_value=temp.name, fail_value=None),
+    ArgOption(cli_name='regexp', init_name='regexp', pass_value=r'\w{2,}', fail_value=r'12('),
     ArgOption(cli_name='mask', init_name='mask', pass_value=temp.name, fail_value=None),
     ArgOption(cli_name='fontfile', init_name='font_path', pass_value=temp.name, fail_value=None),
     ArgOption(cli_name='color', init_name='color_func', pass_value='red', fail_value=None),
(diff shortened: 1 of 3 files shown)


test/test_wordcloud_cli.py
```diff
@@ -24,6 +24,7 @@
 ]
 ARGUMENT_SPEC_REMAINING = [
     ArgOption(cli_name='stopwords', init_name='stopwords', pass_value=temp.name, fail_value=None),
+    ArgOption(cli_name='regexp', init_name='regexp', pass_value=r'\w{2,}', fail_value=r'12('),
```
amueller: Can you explain the fail_value?

amueller:
could you please add a test for this?

test/test_wordcloud_cli.py
boidolr: Sure. The `fail_value` is an invalid regular expression as the opening bracket is never closed. If you have another value in mind just replace it.

boidolr:
I assume you'd like a test for the changed word count?

amueller:
any test that suggest that the regex is used appropriately. You could check the dictionary with different regexps

test/test_wordcloud_cli.py
amueller: oh sorry I misread where this is... never mind, sure, makes sense.

amueller:
thanks :)

PR #296, opened 2017-09-09
Reduce mapping code

boidolr:
Instead of manually adding every new option one can pass the arguments through from the parsing code.
This simplifies adding additional command line options.

### DIFF
--- a/wordcloud/wordcloud_cli.py
+++ b/wordcloud/wordcloud_cli.py
@@ -143,31 +143,29 @@ def parse_args(arguments):
     if args.colormask and args.color:
         raise ValueError('specify either a color mask or a color function')
 
-    with args.text:
-        args.text = args.text.read()
+    args = vars(args)
 
-    if args.stopwords:
-        with args.stopwords:
-            args.stopwords = set(map(str.strip, args.stopwords.readlines()))
+    with args.pop('text') as f:
+        text = f.read()
 
-    if args.mask:
-        args.mask = np.array(Image.open(args.mask))
+    if args['stopwords']:
+        with args.pop('stopwords') as f:
+            args['stopwords'] = set(map(str.strip, f.readlines()))
+
+    if args['mask']:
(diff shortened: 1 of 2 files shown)


wordcloud/wordcloud_cli.py
```diff
@@ -146,10 +140,16 @@ def parse_args(arguments):
...
         color_func = wc.get_single_color_func(args.color)
 
-    args.collocations = not args.no_collocations
-
     args.color_func = color_func
-    return args
+
+    args = vars(args)
+    # remove arguments not directly mapped
+    _ = args.pop('colormask', None)
```
amueller: so these are just ignored? That seems odd.

amueller:
This conflicts with #295, right? So merge this one first and adjust the other one?

amueller:
otherwise looks good

wordcloud/wordcloud_cli.py
boidolr: No - they are used to choose the right color function some lines above and mapped to the same argument for the word cloud instance.

boidolr:
It only conflicts insofar as there is one change in #295 which can be removed after this is merged ;)
I'll happily adapt the other PR.

wordcloud/wordcloud_cli.py
amueller: ah. maybe use ``args.pop('color')`` there so it's more obvious?

boidolr: Right. Just using the `vars(args)` part earlier.

boidolr:
Great I broke it. Will look at it again later and fix. Sorry for the inconvenience.

amueller:
I merged the other one, but I think it'll be easy to fix here.

boidolr:
Done & rebased. No other changes to the commits.

amueller:
great, thanks :)

PR #309, opened 2017-10-19
Catch error if colour image is smaller than canvas

Peter92:
Also add an option to use a default colour instead of raising an error.

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -271,14 +271,14 @@ def test_unicode_stopwords():
 
     assert_true(words_unicode == words_str)
 
+    
 def test_recolor_too_small():
     # check exception is raised when image is too small
-    def _test():
-        colouring = np.array(Image.new('RGB', size=(20, 20)))
-        wc = WordCloud(width=30, height=30).generate(THIS)
-        image_colors = ImageColorGenerator(colouring)
-        wc.recolor(color_func=image_colors)
-    assert_raises_regex(ValueError, 'ImageColorGenerator is smaller than the canvas', _test)
+    colouring = np.array(Image.new('RGB', size=(20, 20)))
+    wc = WordCloud(width=30, height=30).generate(THIS)
+    image_colors = ImageColorGenerator(colouring)
+    assert_raises_regex(ValueError, 'ImageColorGenerator is smaller than the canvas',
+                        wc.recolor, color_func=image_colors)
 
 
 def test_recolor_too_small_set_default():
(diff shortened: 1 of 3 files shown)


amueller:
sweet, thanks! Could you please add a regression test?

Peter92:
That I'm not quite sure how to do haha, is it just running `test/test_wordcloud.py` to check for any errors?

amueller:
There was an error but it looks like it's something on the server going wrong.
What I meant was adding a test to ``test/test_wordcloud.py`` that would have failed before your fix.
For the index error it would just be adding the thing that was failing as a test.
For the error message that would mean inputting the wrong size and doing an ``assert_raises_regex`` http://nullege.com/codes/search/nose.tools.assert_raises_regex to see if the right error is thrown.

Peter92:
That took an awful lot of figuring out as I've never even looked at assert before haha, I think I managed it though

test/test_wordcloud.py
```diff
@@ -268,3 +271,19 @@ def test_unicode_stopwords():
...
     assert_true(words_unicode == words_str)
 
+def test_recolor_too_small():
+    # check exception is raised when image is too small
+    def _test():
+        colouring = np.array(Image.new('RGB', size=(20, 20)))
+        wc = WordCloud(width=30, height=30).generate(THIS)
+        image_colors = ImageColorGenerator(colouring)
+        wc.recolor(color_func=image_colors)
+    assert_raises_regex(ValueError, 'ImageColorGenerator is smaller than the canvas', _test)
```
amueller: you could have just passed "wc.recolor" as the method, I think, but this way is fine, too.

wordcloud/wordcloud.py
```diff
@@ -404,8 +404,11 @@ def generate_from_frequencies(self, frequencies, max_font_size=None):
                                                max_font_size=self.height)
                 # find font sizes
                 sizes = [x[1] for x in self.layout_]
-                font_size = int(2 * sizes[0] * sizes[1] / (sizes[0]
-                                                           + sizes[1]))
+                try:
+                    font_size = int(2 * sizes[0] * sizes[1]
```
amueller: Could you please add a test for this, too? It didn't look like it was included, but I might have missed it.

Peter92:
Good point about passing in the `wc.recolor`, didn't think of that. Also I didn't do the font size one as it seemed to happen kinda randomly, however I just found that the fix didn't work for super small images so I added another exception for it :)

amueller:
awesome, thanks. Merge?

amueller:
awesome, thanks. Merge?

Peter92:
Yeah sure, let me know if I have to do anything on my end haha

amueller:
thanks for your help!

PR #312, opened 2017-10-24
Fix for unicode stopwords files

khamilowicz:
Passing stopwords file with unicode characters caused

```python
TypeError: descriptor 'strip' requires a 'str' object but received a 'unicode'
```


amueller:
Thanks! can you add a regression test, please?

amueller:
fixed in #438

PR #324, opened 2017-11-22
adding one example for understanding generate with frequency

yoonsubKim:


amueller:
Hi. Thanks for the PR. Can you explain what the example illustrates? How to use ``generate_from_frequencies``? Why did you add additional texts and images for that? Also, examples should just be in the examples directory and start with ``plot_`` to be picked up by sphinx-gallery.

yoonsubKim:
Dear
this example illustrates what you exactly said. And due to I thought that example directory is confused, so I made subdirectory. 
I would move `example` directory unless I kept the policy
Lastly, I wish it should be an independent text file, but I will change the way to what you intend.
However, isn't it an unclean `example` directory? please consider that.

amueller:
Well, now you add a single file to the folder, otherwise you'd add a subfolder. How is that more clean? And why would you with to use an independent text file?

amueller:
you're right, the example folder is not very nice right now, we should try to clean it up. But adding more examples is not really the way to do that ;)

yoonsubKim:
the reason why I made an independent text file is that it is easy to explain frequencies. The content of the text is about tag cloud. Therefore, everybody can guess what the main words of the file are such as ‘tag’, ‘cloud’, ‘image’. Just this. The reason is so tiny

yoonsubKim:
So, Is there something I should change more?

yoonsubKim:
well.. I think it is essential to provide how to use `generate_from_frequencies` method. When I made something with `generate_from_frequencies`, I was in trouble to make frequency parameter, so, this is the reason why I made.

examples/frequency.py
```diff
@@ -0,0 +1,48 @@
...
+    for key in tmpDict:
+        fullTermsDict.add(key,tmpDict[key])
+    return fullTermsDict
+
+
+
+def makeImage(text):
+    alice_mask = np.array(Image.open("alice_mask.png"))
+
+    stopwords = {" "}
```
amueller: the stopwords are actually ignored when using ``generate_from_frequencies`` so this is unnecessary.

yoonsubKim: My gosh..
I didn't know that it can work without stopwords. And I fix promptly as fast as I can.

amueller:
ok, looks good.

yoonsubKim:
Pretty nice!, and there is one reason why I've tried to add some codes is just give some appreciation to you about what I can use this Library.

PR #329, opened 2017-12-25
use worldcloud with chinese

FontTian:
Wordcloud with chinese
=======================

Wordcloud is a very good tools, but if you want to create
Chinese wordcloud only wordcloud is not enough. The file
shows how to use wordcloud with Chinese. First, you need a
Chinese word segmentation library jieba, jieba is now the
most elegant the most popular Chinese word segmentation tool in python.
You can use 'PIP install jieba'. To install it. As you can see,
at the same time using wordcloud with jieba very convenient


amueller:
Thank you for the example. Could you maybe create a new subfolder "images" and put the resulting images in it and "resources" and put the stopwords and font in it? The current example folder is a bit messy.
Also, Where did you get the stopwords and font from? Can you maybe add a comment on where the stopwords come from to the python file? And we probably need to mention the license of the font.

FontTian:
I'll finish it as soon as possible.

FontTian:
Maybe now it is ok.If there are any questions on this submission, please let me know,Thank you.

examples/wordcloud_cn.py
```diff
@@ -0,0 +1,95 @@
...
+at the same time using wordcloud with jieba very convenient
+"""
+
+import jieba
+jieba.enable_parallel(4)
+# Setting up parallel processes :4 ,but unable to run on Windows
+from os import path
+from scipy.misc import imread
+import matplotlib.pyplot as plt
+# jieba.load_userdict("txt\userdict.txt")
```
amueller: if this is not needed, please remove it.

examples/wordcloud_cn.py
```diff
@@ -0,0 +1,95 @@
...
+# add userdict by load_userdict()
+from wordcloud import WordCloud, ImageColorGenerator
+
+d = path.dirname(__file__)
+
+stopwords_path = d + '/wc_cn/stopwords_cn_en.txt'
+# Chinese fonts must be set
+font_path = d + '/fonts/SourceHanSerif/SourceHanSerifK-Light.otf'
```
amueller: can you include a Chinese font that's not too big with the PR? Otherwise people will have trouble running the example. Or the example could download the font.

amueller: oh sorry, I just saw it's included, never mind!

examples/wc_cn/README.txt
```diff
@@ -0,0 +1,19 @@
...
+LuXun_black : Thousand pointing fingers, willing ox bow (Zhao Yannian's Woodcut 1961)
+
+LuXun_color : Thousand pointing fingers, willing ox bow (Yin Shoushi's Chinese Ink 1973)(after image processing)
+
+
+According to the copyright law of People's Republic of China, Call to Arms and two works of Artcan be used as a scientific research and personal study, a small amount of copying and translation.
+
+But if you want to use this literary work in other ways, please refer to the copyright law of the People's Republic of China or the relevant laws of the corresponding country.
+
+P.S. Mr. Lu Xun is a spiritual mentor of my life and a good friend of mine. I like him very much.
```
amueller: Whose voice is this here?

examples/wordcloud_cn.py
FontTian: This command allows the user to import a custom word library using the txt file. If you think it's useless, I'll delete it.

examples/wc_cn/README.txt
FontTian: ......I like Lu Xun .but now,I don't think it really fits in here.I'll fix this mistake.

amueller:
thanks!

FontTian:
It is my pleasure to participate in this project.

PR #335, opened 2018-01-26
Add option to reduce generated image size

boidolr:
See documentation here: http://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html
Basically for file types where the option is supported smaller images should be generated if that is possible. For the one's not supporting that flag it is ignored (http://pillow.readthedocs.io/en/stable/reference/Image.html?#PIL.Image.Image.save).


amueller:
thanks :)

PR #336, opened 2018-01-27
Use iterators instead of building complete lists for every step

boidolr:
Speeding up text processing slightly without changing any function interfaces or complicated setup/algorithms.

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -510,7 +510,7 @@ class WordCloud(object):
         include all those things.
         """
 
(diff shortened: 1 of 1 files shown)


amueller:
cool thanks. Any benchmarks? ;)
Sorry for the late reply.

boidolr:
You were right to ask for those - I am embarrased to say that I've not tested them here. While for other code I previously saw speedups and thought why not change it here as well it's actually slower. Sorry!

amueller:
That's slightly surprising, but yeah, you never know :)

PR #338, opened 2018-02-04
Fix error if image contains black chunks

Peter92:
I noticed that a few images were failing to generate, and figured out an issue. If the image contains any large chunks of black, it'll fail, as when it's getting the colour patches, the line `if not reshape.any()` (when `reshape` only contains `0` due to the black) will trigger and assume the image is invalid.

As `reshape` can still be valid even when only containing 0, I switched it to `reshape.size`, where it'll now only flag up if `reshape` is empty. I'm not sure in what case it's supposed to actually trigger though, I attempted with a 1x1 image and it was still alright.

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -270,23 +270,6 @@ def test_unicode_stopwords():
     words_str = wc_str.process_text(str(THIS))
 
     assert_true(words_unicode == words_str)
-
-    
-def test_recolor_too_small():
-    # check exception is raised when image is too small
-    colouring = np.array(Image.new('RGB', size=(20, 20)))
-    wc = WordCloud(width=30, height=30).generate(THIS)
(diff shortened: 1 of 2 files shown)


amueller:
Thanks for the PR and sorry for the slow reply. Check out the failing test. It shows when it should error.
Can you please add a regression test for the issue you found?

Peter92:
Oops, just had a quick look and realised I was the one who added the fix that contained the `reshape.any()` line to begin with. I think that fixed a particular error, but also had to catch some exceptions (unknowingly caused by the fix). I'll try remove the checks sometime and see if it's all fine now, the failing test was just when it was trying to raise the exceptions and the code ended up working fine.

amueller:
Fixed this in #391. Sorry I hadn't been following up.

Peter92:
Alright sure np :)

PR #348, opened 2018-03-05
Add support for mask contour (resolves #46)

### LINKED ISSUE #46: Is there a way to put a border to the shape?

jsmedmar:
Add arguments `contour_width` and `contour_color` used to create a mask contour (see #46). Happy to remove the example, is not as nice as the current one (not sure why the difference in the color palette, see image below).

- ✅ tests included
- 📘 docs updated

Python usage:

```python
wc = WordCloud(mask=mask, contour_width=1, contour_color='steelblue')
```

Bash:

```bash
python -m wordcloud.wordcloud_cli \
    --mask examples/alice_mask.png \
    --text examples/alice.txt \
    --contour_width 20 \
    --contour_color blue \
    --imagefile test.png
```

Here is an example, contours are better with higher quality images:

alice


wordcloud/wordcloud.py
```diff
@@ -679,3 +684,45 @@ def __array__(self):
...
+            boolean_mask = mask == 255
+        elif mask.ndim == 3:
+            # if all channels are white, mask out
+            boolean_mask = np.all(mask[:, :, :3] == 255, axis=-1)
+        else:
+            raise ValueError("Got mask of invalid shape: %s"
+                                % str(mask.shape))
+        return boolean_mask
+
+    def draw_contour(self, img, mask, contour_width=1, contour_color='black'):
```
amueller: Shouldn't this be private? Or why do you want it to be public? If it's public, it needs to document its parameters (even if it's private, it probably should).

test/test_wordcloud.py
```diff
@@ -198,6 +198,38 @@ def test_mask():
...
+    lg_total = lg_array[100:150, 300:400].sum()
+
+    sc = WordCloud(mask=mask, contour_width=1, scale=2, contour_color='blue')
+    sc.generate(THIS)
+    sc_array = np.array(sc)
+    sc_total = sc_array[100:150, 300:400].sum()
+
+    # values were obtained experimentally
+    assert_equal(bk_total, 0)
+    assert_equal(sm_total, 75480)
```
amueller: It's unclear where these values come from and why they test the right thing.

amueller:
Thanks for the PR. I was first a bit skeptical about adding more parameters but it looks like a nice addition. I don't particularly like the tests, though (and haven't reviewed the rest fully).

amueller:
Why not use https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.ndimage.morphology.binary_dilation.html ?

wordcloud/wordcloud.py
```diff
@@ -679,3 +684,45 @@ def __array__(self):
...
+        """Draw mask contour on a pillow image."""
+        mask = self._get_bolean_mask(mask) * 255
+        contour = Image.fromarray(mask.astype(np.uint8))
```
amueller: Can you not do ``mask[:-1] != mask[1:] + mask[:, :-1] != mask[:, 1:]`` or something like that?

jsmedmar:
### Why Not Dilation

Three reasons:

- One of your dependencies (PIL) already solves the `find the edge` problem
- The PR add's the contour feature without adding complex logic into `wordcloud`
- At the moment I don't have more time to put into the project

### Tests

Updated test to:

```python
def test_mask_contour():
    # test mask contour is created
    mask = np.zeros((234, 456), dtype=np.int)
    mask[100:150, 300:400] = 255

    sm = WordCloud(mask=mask, contour_width=1, contour_color='blue')
    sm.generate(THIS)
    sm_array = np.array(sm)
    sm_total = sm_array[100:150, 300:400].sum()

    lg = WordCloud(mask=mask, contour_width=20, contour_color='blue')
    lg.generate(THIS)
    lg_array = np.array(lg)
    lg_total = lg_array[100:150, 300:400].sum()

    sc = WordCloud(mask=mask, contour_width=1, scale=2, contour_color='blue')
    sc.generate(THIS)
    sc_array = np.array(sc)
    sc_total = sc_array[100:150, 300:400].sum()

    # test `contour_width`
    assert_greater(lg_total, sm_total)

    # test contour varies with `scale`
    assert_greater(sc_total, sm_total)

    # test `contour_color`
    assert_equal(sm_array[100, 300], np.array([0, 0, 255]))
```

These tests rely on the fact that contour is drawn inside and outside edge. These examples show the images produced by the test with a red box around the edge for different values of `contour_width`. I assert that sum of blue pixels within red rectangle increase with `contour_width` and `scale`. Lastly, `contour_color` is tested by checking that a pixel within the contour is actually blue.

| `contour_width=0` | `contour_width=10` | `contour_width=20` |
| ----------------- | ------------------ | ------------------ |
| ![box][box]       | ![10][10]          | ![20][20]          |


[box]: https://user-images.githubusercontent.com/8843150/37049604-5d947c1a-213f-11e8-8e1a-fd3331eec0d1.png
[10]: https://user-images.githubusercontent.com/8843150/37049815-1ee8720e-2140-11e8-8de6-45137386da63.png
[20]: https://user-images.githubusercontent.com/8843150/37049835-2f951e40-2140-11e8-9eed-243628a11181.png

wordcloud/wordcloud.py
jsmedmar: I assume you are referring to `_get_boolean_mask`, I didn't write that logic but wanted to reuse it, so put it into a private method `_get_boolean_mask`. I rather not mess with already working code.

wordcloud/wordcloud.py
jsmedmar: Made it private!

amueller:
What I was trying to say was that there are easier methods for edge-finding that require less code using scipy.ndimage. You're right, scipy is not a requirement, so this might be better.

wordcloud/wordcloud.py
amueller: No, I meant instead of calling "FIND_EDGES" which requires converting to an image and then converting back to a numpy array.

jsmedmar:
What did you think about the tests, does it make sense?

jsmedmar:
Guess maybe you want a more pure solution? Let me know if there is anything else I could do

amueller:
thanks.

PR #351, opened 2018-03-07
Avoid relative import in wordcloud_cli.py.

xuhdev:
The relative import in wordcloud_cli.py makes it unusable. The following error always occurs on
Python 3.x:

Parent module '' not loaded, cannot perform relative import

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,4 +16,4 @@ env:
     - DISTRIB="conda" PYTHON_VERSION="3.6" NUMPY_VERSION="1.13.1"
 install: source continuous_integration/install.sh
 cache: apt
-script: nosetests
+script: nosetests && test/test_script.sh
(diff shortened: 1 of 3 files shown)


amueller:
I believe the relative import is correct. How did you install it?

xuhdev:
@amueller `pip3 install --user wordcloud`

After installation, the cli script is located in `bin`, while the rest is in `lib`. I think that's the reason why it's not working.

xuhdev:
Another way to fix this may be to not install wordcloud_cli.py into the bin directory -- instead, adding another small script to bin to call the main function in wordcloud_cli.py.

amueller:
can you add a test that makes sure this works?

xuhdev:
@amueller Test added.

amueller:
thanks!

amueller:
Hm travis failed after merging. Can you look into it?

xuhdev:
This is weird: the testing on the executable failed after the merger. Let me try to fix this.

PR #370, opened 2018-05-04
increase robustness of "wordcloud_cn" in windows

duohappy:
"f_stop_text.split('\n') " is right. Using "splitlines" may be better.


amueller:
thanks.

PR #372, opened 2018-05-05
improve performance of 'wordcloud_cn'

duohappy:
I modify some code to improve performance in Windows, and reduce the possibility of the same words appearing several times.

### DIFF
--- a/examples/wordcloud_cn.py
+++ b/examples/wordcloud_cn.py
@@ -1,4 +1,4 @@
-# - * - coding: utf - 8 -*-
+# -*- coding: utf-8 -*-
 """
 create wordcloud with chinese
 =======================
@@ -12,17 +12,18 @@ You can use 'PIP install jieba'. To install it. As you can see,
 at the same time using wordcloud with jieba very convenient
 """
 
+import os
+
 import jieba
 jieba.enable_parallel(4)
 # Setting up parallel processes :4 ,but unable to run on Windows
-from os import path
 from scipy.misc import imread
 import matplotlib.pyplot as plt
 # jieba.load_userdict("txt\userdict.txt")
 # add userdict by load_userdict()
 from wordcloud import WordCloud, ImageColorGenerator
 
-d = path.dirname(__file__)
+d = os.path.split(os.path.realpath(__file__))[0]  # absolute path of the file
 
 stopwords_path = d + '/wc_cn/stopwords_cn_en.txt'
 # Chinese fonts must be set
@@ -32,10 +33,11 @@ font_path = d + '/fonts/SourceHanSerif/SourceHanSerifK-Light.otf'
 imgname1 = d + '/wc_cn/LuXun.jpg'
 imgname2 = d + '/wc_cn/LuXun_colored.jpg'
 # read the mask / color image taken from
-back_coloring = imread(path.join(d, d + '/wc_cn/LuXun_color.jpg'))
+back_coloring = imread(d + '/wc_cn/LuXun_color.jpg')
 
 # Read the whole text.
-text = open(path.join(d, d + '/wc_cn/CalltoArms.txt')).read()
+with open(d + '/wc_cn/CalltoArms.txt', encoding='utf-8') as f:
+    text = f.read()
 
 # if you want use wordCloud,you need it
 # add userdict by add_word()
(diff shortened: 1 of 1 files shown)


examples/wordcloud_cn.py
```diff
@@ -12,17 +12,18 @@
...
 jieba.enable_parallel(4)
 # Setting up parallel processes :4 ,but unable to run on Windows
-from os import path
 from scipy.misc import imread
 import matplotlib.pyplot as plt
 # jieba.load_userdict("txt\userdict.txt")
 # add userdict by load_userdict()
 from wordcloud import WordCloud, ImageColorGenerator
 
-d = path.dirname(__file__)
```
amueller: Is that not equivalent but just longer?

examples/wordcloud_cn.py
```diff
@@ -32,10 +33,11 @@
 imgname1 = d + '/wc_cn/LuXun.jpg'
 imgname2 = d + '/wc_cn/LuXun_colored.jpg'
 # read the mask / color image taken from
-back_coloring = imread(path.join(d, d + '/wc_cn/LuXun_color.jpg'))
+back_coloring = imread(d + '/wc_cn/LuXun_color.jpg')
```
amueller: this is not the proper way to do this.

examples/wordcloud_cn.py
```diff
@@ -32,10 +33,11 @@
...
 
 # Read the whole text.
-text = open(path.join(d, d + '/wc_cn/CalltoArms.txt')).read()
+with open(d + '/wc_cn/CalltoArms.txt', encoding='utf-8') as f:
```
amueller: same here.

examples/wordcloud_cn.py
```diff
@@ -60,9 +62,10 @@ def jieba_processing_txt(text):
             mywordlist.append(myword)
     return ' '.join(mywordlist)
 
-
+# the return value of 'jieba_processing_txt' function is words, which have been cleaned
+# collocations=False, the parameter is necessary to reduce the possibility of same words appearing several times
 wc = WordCloud(font_path=font_path, background_color="white", max_words=2000, mask=back_coloring,
-               max_font_size=100, random_state=42, width=1000, height=860, margin=2,)
+               max_font_size=100, random_state=42, width=1000, height=860, margin=2, collocations=False)
```
amueller: this seems good here.

examples/wordcloud_cn.py
amueller: above it's using ``d + .. `` it should be using ``os.path.join`` everywhere.

examples/wordcloud_cn.py
duohappy: In my Windows and Linux, 'path.dirname(__file__)' does't work.

``` python
# -*- coding: utf-8 -*-
#  test.py
import os

d = os.path.dirname(__file__)
print(d)
```
touch a file named test.py in your home directory, '/home/yourhome'.
if you run the python file in your home directory, like 'python test.py', print nothing
if you run the python file ,like 'python /home/yourhome/test.py', print '/home/yourhome'

the return value is not certain, which is confusing. 'os.path.split(os.path.realpath(__file__))[0]' always is fine.

examples/wordcloud_cn.py
duohappy: 'path.join' is proper, and 'encoding='utf-8' is necessary.
In python help file, the parameter is platform dependent. If I run the python file in Windows with out 'encoding='utf-8', return error
> encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever locale.getpreferredencoding() returns)

examples/wordcloud_cn.py
duohappy: yes, adding 'collocations=False' maybe better.
in the python file, some words is deleted, which maybe cause collocations.
``` python
for myword in liststr.split('/'):
        if not (myword.strip() in f_stop_seg_list) and len(myword.strip()) > 1:
            mywordlist.append(myword)
```

examples/wordcloud_cn.py
amueller: So do you want to change to join again?

duohappy: yes, 'path.join' is better

Issue #371, opened 2018-05-05
ValueError: cannot convert float NaN to integer


hossein-pourbozorg:
>>> import wordcloud
>>> import math
>>> wc = wordcloud.WordCloud()
>>> wc.generate_from_frequencies({'a': math.nan})
Traceback (most recent call last):
  File "<pyshell#4>", line 1, in <module>
    wc.generate_from_frequencies({'a': math.nan})
  File "/usr/local/lib/python3.6/dist-packages/wordcloud/wordcloud.py", line 431, in generate_from_frequencies
    + (1 - rs)) * font_size))
ValueError: cannot convert float NaN to integer
>>> wc.generate_from_frequencies({'a': math.inf})
Traceback (most recent call last):
  File "<pyshell#5>", line 1, in <module>
    wc.generate_from_frequencies({'a': math.inf})
  File "/usr/local/lib/python3.6/dist-packages/wordcloud/wordcloud.py", line 431, in generate_from_frequencies
    + (1 - rs)) * font_size))
ValueError: cannot convert float NaN to integer>>> import wordcloud
>>> import math
>>> wc = wordcloud.WordCloud()
>>> wc.generate_from_frequencies({'a': math.nan})
Traceback (most recent call last):
  File "<pyshell#4>", line 1, in <module>
    wc.generate_from_frequencies({'a': math.nan})
  File "/usr/local/lib/python3.6/dist-packages/wordcloud/wordcloud.py", line 431, in generate_from_frequencies
    + (1 - rs)) * font_size))
ValueError: cannot convert float NaN to integer
>>> wc.generate_from_frequencies({'a': math.inf})
Traceback (most recent call last):
  File "<pyshell#5>", line 1, in <module>
    wc.generate_from_frequencies({'a': math.inf})
  File "/usr/local/lib/python3.6/dist-packages/wordcloud/wordcloud.py", line 431, in generate_from_frequencies
    + (1 - rs)) * font_size))
ValueError: cannot convert float NaN to integer

amueller:
Why should this work, and what should it do?

hossein-pourbozorg:
if this shouldn't work.
i think, it's better to raise a more explicit error.
it seem like a bug to me at the first time.

PR #373, opened 2018-05-05
bugfix: no need for int when use round

hossein-pourbozorg:
round function with no ndigits argument return int [1]
so, no need for int when use round

#371 

refs:
[1] https://docs.python.org/3/library/functions.html#round

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -426,8 +426,8 @@ class WordCloud(object):
             # select the font size
             rs = self.relative_scaling
             if rs != 0:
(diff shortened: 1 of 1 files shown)


hossein-pourbozorg:
it's better to use math.isfinite function [1] to check this kind of cases

refs:
[1] https://docs.python.org/3/library/math.html#math.isfinite

amueller:
I'm not sure what this has to do with #371.

hossein-pourbozorg:
i checked more and see this change doesn't fix error
so, i close the pull request.

PR #375, opened 2018-05-11
fix some of pylint warnings

hossein-pourbozorg:


doc/conf.py
```diff
@@ -59,7 +59,7 @@
     'examples_dirs': '../examples',
     # path where to save gallery generated examples
     'gallery_dirs': 'auto_examples',
-    'filename_pattern': '/\w+'}
+    'filename_pattern': r'/\w+'}
```
amueller: That changes the semantics, right? also that is the conf of the doc....

wordcloud/wordcloud.py
```diff
@@ -147,7 +149,7 @@ def single_color_func(word=None, font_size=None, position=None,
 
 
 class WordCloud(object):
-    """Word cloud object for generating and drawing.
+    r"""Word cloud object for generating and drawing.
```
amueller: That doesn't change anything, does it?

wordcloud/wordcloud_cli.py
```diff
@@ -12,12 +12,15 @@
 import io
 import re
 import argparse
-import wordcloud as wc
-import numpy as np
+
 from PIL import Image
 
+import numpy as np
+
```
amueller: I feel like adding so many newlines decreases readability. What linter asks that?

doc/conf.py
hossein-pourbozorg: this string contains backslash character. so, we should use "r" prefix for treat backslashes as literal characters

wordcloud/wordcloud.py
hossein-pourbozorg: this string contains backslash character. so, we should use "r" prefix for treat backslashes as literal characters
(in second line of regexp parameter section, you use "r" prefix but it doesn't do anything because its in a string, so we should use "r" prefix for parent string)

doc/conf.py
amueller: ooohh for some reason I thought it was escaped but it's not... you're right.

wordcloud/wordcloud_cli.py
hossein-pourbozorg: pep8 says:
```
You should put a blank line between each group of imports.
```

https://www.python.org/dev/peps/pep-0008/#imports

amueller: pe8 mentions three groups, though. and there are 9 here now, right?

hossein-pourbozorg: yes,
i seperate "import ..." statement and "from ... import ..." statement for show the difference between modules and functions or variables clearly
in my opinion, it's better to use only "from ... import ..." statement to know what functions or variables used and when an api changed, you can just use easily "from ... import ... as ..." statement

jcfr:
Closing. This is superseded by #415

PR #389, opened 2018-06-05
Fixed bug for generating images with transparent background.

m000:
I had problems generating images with `mode='RGBA', background_color=None`. The contour used when rendering the image was lacking a transparency channel. This should fix it.


amueller:
looks good, can you add a test?

m000:
Should be good now.

test/test_wordcloud.py
```diff
@@ -229,6 +230,18 @@ def test_mask_contour():
...
 
+def test_mask_contour_transparent():
+    # test if mask contour can be drawn on a transparent background
+    # https://github.com/amueller/word_cloud/pull/389
+    examples_dir = os.path.realpath(os.path.join(os.path.dirname(__file__), '..', 'examples'))
+    mask_file = os.path.join(examples_dir, 'alice_mask.png')
+    mask = np.array(Image.open(mask_file))
+    w = WordCloud(mode='RGBA', background_color=None, contour_color='blue', contour_width=2, mask=mask,
+                  max_words=50, width=800, height=800)
+    w.generate(THIS)
```
amueller: So does this give an error? You could probaby also just create an empty mask to reproduce it instead of loading one, right?

test/test_wordcloud.py
```diff
@@ -229,6 +230,18 @@ def test_mask_contour():
...
+    w.to_file(NamedTemporaryFile(suffix=".png").name)
```
amueller: Why is this necessary?

test/test_wordcloud.py
m000: I don't know what you describe works, but why would you be trying to draw a contour with an empty mask?

amueller: Right maybe that's not the very best idea but you could just add a stripe to one side or something. The point is to make the test as simple and self-contained as possible. And if you can reproduce the error with a small numpy array that would be easier.

test/test_wordcloud.py
m000: That's how the previously buggy code path is triggered. If you don't write to a file the contour is not drawn. The exception happens here:
```
ret = np.array(img) * np.invert(contour)
ValueError: operands could not be broadcast together with shapes (900,900,4) (900,900,3)
```

amueller: you could just do ``to_image()``.

m000: Ok. Fixed that.

test/test_wordcloud.py
m000: The image is in the repo, so it is adequately self-contained. Sure, someone with a better grasp of PIL can remove this dependency in the future.

amueller:
```python
mask = np.ones((100, 100))
mask[:, -10:] = 0
```
should do it, right?

amueller: Actually better
```
mask = np.zeros((100, 100, 3), dtype="ubyte")
mask[:, -10:] = 255
```

m000: Yup! Looks good. This would have taken me at least another 30' to figure out. Haven't worked with matrices and stuff for quite a long time.

amueller:
can you remind me what the reason was to change the default color?

VergeDX:
I rebase the patches to the currently stable branch `1.8.1`:

```````patch
diff --git a/wordcloud/wordcloud.py b/wordcloud/wordcloud.py
index 26b0d54..93dc6b1 100644
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -311,7 +311,7 @@ class WordCloud(object):
                  max_font_size=None, font_step=1, mode="RGB",
                  relative_scaling='auto', regexp=None, collocations=True,
                  colormap=None, normalize_plurals=True, contour_width=0,
-                 contour_color='black', repeat=False,
+                 contour_color=None, repeat=False,
                  include_numbers=False, min_word_length=0, collocation_threshold=30):
         if font_path is None:
             font_path = FONT_PATH
@@ -1006,7 +1006,7 @@ class WordCloud(object):
 
     def _draw_contour(self, img):
         """Draw mask contour on a pillow image."""
-        if self.mask is None or self.contour_width == 0:
+        if self.mask is None or self.contour_color is None or self.contour_width == 0:
             return img
 
         mask = self._get_bolean_mask(self.mask) * 255
@@ -1024,12 +1024,14 @@ class WordCloud(object):
         contour = Image.fromarray(contour)
         contour = contour.filter(ImageFilter.GaussianBlur(radius=radius))
         contour = np.array(contour) > 0
-        contour = np.dstack((contour, contour, contour))
+        if img.mode == 'RGBA':
+            contour = np.dstack((contour, contour, contour, contour))
+        else:
+            contour = np.dstack((contour, contour, contour))
 
         # color the contour
         ret = np.array(img) * np.invert(contour)
-        if self.contour_color != 'black':
-            color = Image.new(img.mode, img.size, self.contour_color)
-            ret += np.array(color) * contour
+        color = np.array(Image.new(img.mode, img.size, self.contour_color))
+        ret += color * contour
 
         return Image.fromarray(ret)
diff --git a/wordcloud/wordcloud_cli.py b/wordcloud/wordcloud_cli.py
index 125a86f..03c7db6 100644
--- a/wordcloud/wordcloud_cli.py
+++ b/wordcloud/wordcloud_cli.py
@@ -126,7 +126,7 @@ def make_parser():
         dest='contour_width',
         help='if greater than 0, draw mask contour (default: 0)')
     parser.add_argument(
-        '--contour_color', metavar='color', default='black', type=str,
+        '--contour_color', metavar='color', default=None, type=str,
         dest='contour_color',
         help='use given color as mask contour color -'
              ' accepts any value from PIL.ImageColor.getcolor')
```````

Leo310:
when will this be merged?

orbsmiv:
Just adding another bump to this bug fix :-)

Issue #381, opened 2018-05-27
Distributable .exe that imports WordCloud fails because of missing file


Michael-E-Rose:
#### Description
I compiled a python3.5 script that includes WordCloud using pyinstaller on Windows 10 in order to have a distributable exe file. The exe fails because it is looking for a file stopwords in folder Wordcloud in the user's temp directory. The script runs all fine when exectued in script or bash mode. The file is being looked up in a folder with cryptic name, which changes on every run. The foldername always starts with _MEI followed by some numbers.

#### Steps/Code to Reproduce
Compile python script on windows with pyinstaller and include line ``from wordcloud import WordCloud`

#### Error message
```
C:\Users\w6504hk\Downloads\Oblaka\dist>Oblaka
Traceback (most recent call last):
  File "Oblaka.py", line 18, in <module>
  File "<frozen importlib._bootstrap>", line 968, in _find_and_load
  File "<frozen importlib._bootstrap>", line 957, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "c:\program files\python35\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "site-packages\wordcloud\__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 968, in _find_and_load
  File "<frozen importlib._bootstrap>", line 957, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "c:\program files\python35\lib\site-packages\PyInstaller\loader\pyimod03_importers.py", line 631, in exec_module
    exec(bytecode, module.__dict__)
  File "site-packages\wordcloud\wordcloud.py", line 29, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\w6504hk\\AppData\\Local\\Temp\\_MEI52242\\wordcloud\\stopwords'
```
I think this behaviour is not intended. It is related to wordcloud or might it be caused by pyinstaller?

amueller:
There is a file called stopwords in wordcloud that it reads It seems pyinstaller doesn't bundle that correctly. A simple but hacky way to work around this would be to include the stopwords in the py file.

Michael-E-Rose:
That's correct, it doesn't bundle that.  The question for me however is: Why does wordcloud require this file, if I use my own precompiled words?

[...]

PR #393, opened 2018-06-10
Load stopwords only when needed

Michael-E-Rose:
Related to #381.  Load the stopwords file only when needed, i.e. when the user does not provide its own stopwords.

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -27,7 +27,6 @@ from .tokenization import unigrams_and_bigrams, process_tokens
 
 FILE = os.path.dirname(__file__)
 FONT_PATH = os.environ.get('FONT_PATH', os.path.join(FILE, 'DroidSansMono.ttf'))
-STOPWORDS = set(map(str.strip, open(os.path.join(FILE, 'stopwords')).readlines()))
 
 
 class IntegralOccupancyMap(object):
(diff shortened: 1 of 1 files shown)


wordcloud/wordcloud.py
```diff
@@ -302,7 +301,11 @@ def __init__(self, font_path=None, width=400, height=200, margin=2,
         self.scale = scale
         self.color_func = color_func or colormap_color_func(colormap)
         self.max_words = max_words
-        self.stopwords = stopwords if stopwords is not None else STOPWORDS
+        if stopwords is None:
+            fname = open(os.path.join(FILE, 'stopwords')
```
amueller: missing closing parenthesis

Michael-E-Rose:
Sorry, my bad.

I just notice that `STOPWORDS` is being imported in `__init__.py` as well and it's used in some of the examples.  While it's no problem to remove them from `__init__.py`, more needs to be done in the examples, because there you add custom words.
There's two alternatives:

1. Read your stopwords file manually and add custom words
2. Use nltk's stopwords (in the examples) and add custom words.

amueller:
Actually I think removing them from ``__init__.py`` is the much bigger issue as it's a backward incompatible change. Why would you want to remove it from the examples?

Michael-E-Rose:
When stopwords are only loaded when `WordCloud(stopwords=None)`, then `.STOPWORDS` doesn't exist. So it can't be imported from the module.  `from wordcloud import STOPWORDS` would fail.

amueller:
Sure. I guess I didn't understand what you mean by removing them from the
example. You meant not using this exact important.
I don't really see the benefit of changing the API for stopwords. The only
issue seems to be that it triggers a bug in .exe generation. So maybe fix
that instead?

amueller:
you can just add the stopwords to pyinstaller: https://pyinstaller.readthedocs.io/en/v3.3.1/spec-files.html#adding-files-to-the-bundle

PR #399, opened 2018-06-21
Add documentation for single color word cloud

### LINKED ISSUE #185: Simple way to include one single color without any function
Hi,

maybe I missed it, but is there an easy way to set the color for every word without using any extra functions? I know that the algorithm you use for word placement is quite different, but I'm basically looking for the easiest way to make the wordclouds look similar to those produced in R. An example can be found at http://datascienceplus.com/building-wordclouds-in-r/:

Thanks =)

maifeng:
Fixed #185.


amueller:
thanks!

PR #406, opened 2018-07-24
circleci: Fix workspace issue avoiding to generate deprecated egg

jcfr:
Using "pip install -e" is sufficient to build in place and allow the
test to work as expected.

This commit fixes the following error:

Concurrent upstream jobs persisted the same file(s) into the workspace:
  - dist/wordcloud-1.4.1-py2.7-linux-x86_64.egg


amueller:
Thanks :)

PR #407, opened 2018-07-25
Add infrastructure for automatic deployment of the website

jcfr:


doc/_static/theme_override.css
```diff
@@ -0,0 +1,7 @@
+div[class^="highlight"] a {
```
amueller: where did this come from?

jcfr: Blaming the file should help answer the question. We could also add a comment to the CSS.

https://github.com/amueller/word_cloud/pull/407/commits/e452ab3562c3317a6eefbee8aa03035d2bd9d0a2

amueller: ahh nice. I'm just redoing the theme a bit (though I really should be doing other things lol).

PR #408, opened 2018-07-25
use read the docs theme on website

amueller:
also add documentation for CLI.


jcfr:
Nitpick:

To help keep the history clean and easy to review, I suggest to avoid merging master into your topic (I usually keep this for long standing topic).

In that particular case, it may be sensible to squash all commit together.

amueller:
I usually squash and merge all PRs, in which case this works nicely. The problem with rebasing PRs is that github will lose all review comments so I'm not usually doing this any more.

jcfr:
> The problem with rebasing PRs is that github will lose all review comments so I'm not usually doing this any more.

I see. There is definitely no general rule. 


That said, in that particular case, I would still do:

```
git checkout -b rtd_theme-squashed origin/rtd_theme 
git reset --soft origin/master
```

And then do one or two commits summarizing the changes and force push.

Then commit with a messages like these ones:

```
doc: Switch to rtd_theme

The existing theme was based of xxxx and had few problems like X, Y, Z. 
By switching to rtd_theme, ... 
```

```
doc: document command line interface introducing sphinx-argparse extension
```

Then `git blame` would behave nicely and allow to more clearly understand the change.

amueller:
you'd have the same benefits if you merge master and then squash and merge in github, though. If you actually provide a good commit message on github ;)

jcfr:
Warnings to fix:

```
/home/jcfr/Projects/word_cloud/doc/cli.rst:4: WARNING: Module "wordcloud.wordcloud_cli" has no attribute "make_parser"
Incorrect argparse :module: or :func: values?
```

PR #421, opened 2018-07-26
cli: Install executable using entry_points

### LINKED ISSUE #420: cli: Add support for installing and running"wordcloud_cli" on all platforms
#### Description

After installing wordcloud as an editable install, it is not possible to run the `wordcloud_cli.py` script available in the source tree.

#### Steps/Code to Reproduce

After installing wordcloud locally using:

```
pip install -e .
```

and then trying to run the CLI with:

```
$ python ./wordcloud/wordcloud_cli.py --text examples/constitution.txt   --imagefile consitution.png
```

It fails with 

```pytb
  File "./wordcloud/wordcloud_cli.py", line 15, in <module>
    import wordcloud as wc
  File "/home/jcfr/Projects/word_cloud/wordcloud/wordcloud.py", line 25, in <module>
    from .query_integral_image import query_integral_image
SystemError: Parent module '' not loaded, cannot perform relative import
```

Note that running the installed cli works:

```
$ wordcloud_cli.py --text examples/constitution.txt   --imagefile consitution.png
```

But this is unlikely to work on windows and prevent developer from improving the CLI without having to reinstall it after each update, or having to  copy the script our and back into the right folder.

jcfr:
This allows to have a consistent way of execution the CLI on all
platforms. On windows, an executable launcher is automatically created
after installing the package:

* https://packaging.python.org/guides/distributing-packages-using-setuptools/#console-scripts
* https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation

This commit also makes it possible for developer to modify the `wordcloud/wordcloud_cli.py`
script after doing an editable install and ensure the updates are considered
without having to re-install.

After downloading the source, if a user tries to directly execute `wordcloud/worcloud_cli.py`,
instead of getting the error message:

```
  SystemError: Parent module '' not loaded, cannot perform relative import
```

it now shows an informative message suggesting how to run the CLI:

```
  To execute the CLI, instead consider running:

    wordcloud_cli --help

  or

    python -m wordcloud --help
```

Fixes #420

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,17 +60,17 @@ Getting fancy with some colors:
 
 ## Command-line usage
 
-The `wordcloud_cli.py` tool can be used to generate word clouds directly from the command-line:
+The `wordcloud_cli` tool can be used to generate word clouds directly from the command-line:
 
-	$ wordcloud_cli.py --text mytext.txt --imagefile wordcloud.png
+	$ wordcloud_cli --text mytext.txt --imagefile wordcloud.png
 
 If you're dealing with PDF files, then `pdftotext`, included by default with many Linux distribution, comes in handy:
 
-	$ pdftotext mydocument.pdf - | wordcloud_cli.py --imagefile wordcloud.png
+	$ pdftotext mydocument.pdf - | wordcloud_cli --imagefile wordcloud.png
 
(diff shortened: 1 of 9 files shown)


amueller:
thanks! how did you find this?

jcfr:
After working on few other large and small python projects, I got familiar with this type of mechanism. :smile:

PR #424, opened 2018-07-26
setup: use scikit-build for building C module

jcfr:
scikit-build is a drop-in replacement to setuptools.setup function
allowing to easily compile and package extensions (C/C++/Cython) by
bridging CMake and setuptools. See http://scikit-build.org

CMake is an open-source, cross-platform family of tools designed
to build, test and package software. See https://cmake.org

Ninja is a small build system with a focus on speed. It differs from other
build systems in two major respects: it is designed to have its input
files generated by a higher-level build system, and it is designed
to run builds as fast as possible. See https://ninja-build.org/

By adding the "pyproject.toml" file, we ensure the required packages
are installed when running "pip wheel . -w dist".

To streamline testing and creation of source or built distribution, the
"requirements-dev.txt" was also updated to include scikit-build, cmake
and ninja.

This commit introduces new build dependencies all available as wheels:
scikit-built, cmake and ninja.


amueller:
I feel like using CMake for this is a bit overkill. Like the configuration and dependencies got significantly longer. Also it requires new contributors to understand more things.

jcfr:
I anticipate, It will streamline the process:
* no need to manually regenerate the C source
  * generating the source given the pyx across all version of pythons just worked on the CI
* if the user want to build from source and doesn't have the compiler environment, scikit-build will give clear message about what to install and where to download it. (e.g version of visual studio, XCode, ... )

There was a lot of message in the issue tracker about which compiler, etc ... this could help. Installing, these wheels is fast and their available for all python version.

jcfr:
Also, if this causes more problem that is is claiming to solve, I will help revert it. As you noticed, I have a good understanding of the word_cloud code base and can definitely help,

amueller:
Alright, feel free to merge.

PR #429, opened 2018-09-10
Replace Jieba with HanLP.

FontTian:
This time, new Chinese word cloud generation method is added, which can effectively improve performance.

pyhanlp is One of the most powerful natural language processing libraries in Chinese today, 
and it's extremely easy to use.You can use 'PIP install pyhanlp'. 
To install it,like Jieba.

Its level of identity of named entity,word segmentation was better than jieba,
and has more ways to do it.You'll save a lot of time when you use it.

And thanks to its excellent performance.
when we handle a large number of Chinese texts
We don't have to use the User-defined dictionaries.


amueller:
Thanks! There's some issue with your pull request, though, you changed way to many files!

PR #441, opened 2018-10-12
Drawing mask contour on a transparent background

thehappydinoa:
This is #389 with conflicts


amueller:
can you remind me what the reason was to change the default color?

amueller:
Maybe @jcfr can help?

jcfr:
> >  I am not sure exactly what went wrong.

> Maybe @jcfr can help?

This should be fixed in https://github.com/amueller/word_cloud/pull/442

PR #465, opened 2019-01-27
Allowing: Minimum Word Length & Frequencies of Numbers

the-rccg:
Added parameters to WordCloud() to allow for filtering frequencies of numbers, as well as filtering by minimum length of a word. 
Numbers are useful for visualizing patterns in numbers, dummy example using random numbers and the first 10,000 digits of pi.
Minimum word length is a simple way for filtering out functional words in foreign languages without having to write stopword lists or use pre-defined functional word collections. Examples filters A New Hope with length rather than stopwords. 
Note: obviously not the most efficient implementations, but kept these in line with the style of the rest of the package.


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
amueller: can you please remove this file?

amueller:
I would maybe not add two more examples as there already are quite a few. Can you maybe add one and also add tests for the two new options? Thanks!

the-rccg:
Will do a clean pull request again, with just one commit and remove the files. Sorry about that.

amueller:
looks good, thanks!

PR #466, opened 2019-01-28
Example: Don't assign variable name back to wordcloud

shuttle1987:
Improve example code to not overwrite the `wordcloud` name


amueller:
Thanks. It's not really overwriting anything because the name wordcloud is not defined in that scope, but you're right, not using it is more clear.

shuttle1987:
I was using this library today in a workshop from within jupyter, someone copied and pasted from the example and this caused some confusion when later code broke. Thanks for the great library!

amueller:
Yeah that makes sense :) thanks for the PR.

PR #467, opened 2019-01-30
Require matplotlib.

Jessime:
* Add matplotlib to install_requires.
* Move matplotlib install to top of wordcloud.py


amueller:
thanks!

Jessime:
You're welcome! Any idea when you're thinking to do a new package release?

amueller:
Not really, I've been very busy with other stuff and there haven't been a lot of changes. Do you need one for something? If so, why? It's not a lot of work (though I won't be able to do it before end of Feb probably).

PR #490, opened 2019-07-03
Update stopwords

kalpit181:


amueller:
Thanks for the PR. why did you change the project URL?

You can change the stopwords when you're using the wordcloud and I think that makes more sense than trying to curate a stopword list that works for every application, because that's not possible.
"will" for example is not a stopword in all contexts (it can mean resolve).
You can just add it to the set of built-in stopwords in the code before calling WordCloud like I did here:

https://github.com/amueller/word_cloud/blob/32f5d2adf8e5f262e3f22757f019492f137a922b/examples/a_new_hope.py#L41

kalpit181:
I am just a newbie at github. Sorry for the inconvenience.

amueller:
no worries, we all started somewhere :)

Issue #58, opened 2015-04-29
Vector graphics


chunrong:
Can a Vector graph be generated? When it is zoomed out, it will be not that clear.
Thanks,

amueller:
The way PIL renders, there are no vector graphics currently. The best way would be to do a HTML export, I think, which is not currently there. There is some code at #23, but it is not really polished or tested.

amueller:
Thinking about it, it might be easier to go via matplotlib. HTML export seems hard to do.

SalvaEB:
There is a word cloud generator based on cairo which already produces vector output (svg and pdf). Although it is written in c++ and not in python, you could take a look just in case: https://github.com/SalvaEB/mask_word_cloud (disclaimer: I am the author. After looking for a word cloud generator that produced vector output, I tried to add this feature to word_cloud and I finally finished by writing a new code in c++ from scratch).

amueller:
thanks for sharing @SalvaEB. I think there are many implementations that do support vector graphics. In particular I'm sure the d3 versions will do.

amueller:
@SalvaEB you use a scanline approach, right? So the algorithm is slightly different but gives the same result? Is it any faster than mine?
And what do you dilate the image for?

SalvaEB:
I use a scanline instead of an integral matrix (I store the number of white pixels until the rightmost left pixel). I also use other tricks to update when writing a new word and others to avoid traversing the full image. I dilate the image in order to avoid a margin around the mask. In this way, the shape of the resulting word cloud seems more accurate. It seems faster but, since it is written in C++, I couldn't say which part is due to algorithmic differences. When using the the minimal font size higher than 4 the algorithm is much faster (and the result is less dense). Best regards, S.

amueller:
Well, mine is written in Cython, which should be as fast basically. I think my current approach is a bit silly and I should probably change it. I'll check out yours.

amueller:
What other tricks do you use?

SalvaEB:
reservoir sampling in order to avoid detecting the valid bounding boxes twice. You can check the code. The treatment of words and words sizes is very dummy for the moment. Unfortunately I have no much time now to complete it.

amueller:
yeah I should also be working on other things ^^

AndreasS2501:
Hi,

Just wanted to check on this. Is it now possible to also generate SVG?

SalvaEB:
Hi,
[...]

PR #492, opened 2019-08-17
Add Bounding Box Scaling Method

Will-Renius:
I added bounding boxes for word sizes to the layout and used these to adjust font size so they don't scale to overlap other words.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -221,6 +221,3 @@ dmypy.json
 # End of https://www.gitignore.io/api/pycharm,python,visualstudiocode
 
 # Custom rules (everything added below won't be overriden by 'Generate .gitignore File' if you use 'Update' option)
-
-scratch
-.vscode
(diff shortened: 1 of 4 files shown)


Will-Renius:
Any interest in merging this feature? I was hoping to get improved scaling into the source code rather than maintaining it separately for my applications. I'm open to other ideas for fixes.

amueller:
Sorry I was busy and didn't see this. Can you explain what the goal is?

Will-Renius:
The goal is to prevent overlapping words when using scale. 

The current scaling method uses the same value to adjust image size and font size. However, fonts don't scale at 1:1 with pixel width. This change keeps the original bounding box used to place the word, scales that, then limits the font to stay within the new bounding box.

amueller:
hm that makes sense.
Would that be solved better by not using scaling and export vector graphics directly, see #58

Will-Renius:
Hard to say, my goal is to keep image size low when computing and then increase resolution in order to speed up processing, which is exactly what the scaling mechanism is there for. If a proper vector graphic implementation is implemented would you want to replace the current scaling mechanism with one that relies on conversion of vector graphics to a PIL image?

amueller:
Yes that's what scaling is for, but it's a bit of a hack. If we have vector graphics support, why would you convert to a PIL image? You could convert the vector graphic to PNG in the end if you wanted to, I guess.

PR #519, opened 2019-12-24
Add to_svg

jojolebarjos:
This is another tentative at vector file export. I initially did some experiments using [freetype-py](https://pypi.org/project/freetype-py/), but I was able to get the same results using Pillow bindings of FreeType (there are [inconsistencies](https://stackoverflow.com/questions/43060479/how-to-get-the-font-pixel-height-using-pil-imagefont), though).

There are however slight mismatches between rasterized and vector outputs. I also spotted inconsistencies using different browsers for some fonts (e.g. Segoe Script). I added an option (`embed_image`) to include the output of `to_image` in the SVG file, which is useful for debugging. I believe most differences are due to subtleties in kerning and ligatures, which is not significant in common fonts.

It may also be useful to consider embedding the font file directly in the SVG file (or only the required subset, ideally). As my personal use case does not require fancy fonts, I did not dig any further on this topic. This means that using the default font will render properly only if Droid Sans Mono is available to the SVG reader.

I have attached [two modified examples](https://github.com/amueller/word_cloud/files/3999068/examples.zip), which use different fonts for portability.

Also, contour is not implemented.

Last but not least, I want to credit @jnothman and [his pull request](https://github.com/amueller/word_cloud/pull/163) (in particular, his hack to detect font name/weight/style). As I am not fluent in forks and pull requests, I had to create a new pull request. But I would be glad to learn how to extend an existing pull request!

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -732,7 +732,7 @@ class WordCloud(object):
     def to_html(self):
         raise NotImplementedError("FIXME!!!")
 
-    def to_svg(self, embed_image=False, embed_font=False):
+    def to_svg(self, embed_font=False, optimize_embedded_font=True, embed_image=False):
         """Export to SVG.
 
         Font is assumed to be available to the SVG reader. Otherwise, text
@@ -749,12 +749,17 @@ class WordCloud(object):
 
         Parameters
         ----------
+        embed_font : bool, default=False
+            Whether to include font inside resulting SVG file.
+
+        optimize_embedded_font : bool, default=True
+            Whether to be aggressive when embedding a font, to reduce size. In
+            particular, hinting tables are dropped, which may introduces slight
+            changes to character shapes (w.r.t. `to_image` baseline).
+
         embed_image : bool, default=False
             Whether to include rasterized image inside resulting SVG file.
             Useful for debugging.
-
-        embed_font : bool, default=False
-            Whether to include font inside resulting SVG file.
 
         Returns
         -------
@@ -825,10 +830,10 @@ class WordCloud(object):
             options = fontTools.subset.Options(
 
                 # Small impact on character shapes, but reduce size a lot
-                hinting=False,
+                hinting=not optimize_embedded_font,
 
                 # On small subsets, can improve size
-                desubroutinize=True,
+                desubroutinize=optimize_embedded_font,
 
                 # Try to be lenient
                 ignore_missing_glyphs=True,
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -9,8 +9,10 @@ from __future__ import division
 
 import warnings
 from random import Random
+import io
 import os
 import re
+import base64
 import sys
 import colorsys
 import matplotlib
@@ -729,7 +731,7 @@ class WordCloud(object):
     def to_html(self):
         raise NotImplementedError("FIXME!!!")
     
-    def to_svg(self):
+    def to_svg(self, embed_image=False):
         """Export to SVG.
 
         Returns
@@ -796,6 +798,21 @@ class WordCloud(object):
                 '>'
                 '</rect>'
                 .format(self.background_color)
+            )
+        
+        # Embed image, useful for debug purpose
+        if embed_image:
+            image = self.to_image()
+            data = io.BytesIO()
+            image.save(data, format='JPEG')
+            data = base64.b64encode(data.getbuffer()).decode('ascii')
+            result.append(
+                '<image'
+                ' width="100%"'
+                ' height="100%"'
+                ' href="data:image/jpg;base64,{}"'
+                '/>'
+                .format(data)
             )
 
         # For each word in layout
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -742,13 +742,12 @@ class WordCloud(object):
         self._check_generated()
 
         # Get output size, in pixels
-        # TODO check self.scale
         if self.mask is not None:
             width = self.mask.shape[1]
             height = self.mask.shape[0]
         else:
             height, width = self.height, self.width
-        
+
         # Get max font size
         if self.max_font_size is None:
             max_font_size = max(w[1] for w in self.layout_)
@@ -760,7 +759,7 @@ class WordCloud(object):
 
         # Prepare global style
         style = {}
-        font = ImageFont.truetype(self.font_path, int(max_font_size))
+        font = ImageFont.truetype(self.font_path, int(max_font_size * self.scale))
         font_family, font_style = font.getname()
         # TODO properly escape/quote this
         # TODO should add option to specify URL for font (i.e. WOFF file)
@@ -783,7 +782,7 @@ class WordCloud(object):
             ' height="{}"'
             ' style="{}"'
             '>'
-            .format(width, height, style)
+            .format(width * self.scale, height * self.scale, style)
         )
 
         # Add background
@@ -800,12 +799,14 @@ class WordCloud(object):
 
         # For each word in layout
         for (word, count), font_size, (y, x), orientation, color in self.layout_:
+            x *= self.scale
+            y *= self.scale
 
             # Get text metrics
-            font = ImageFont.truetype(self.font_path, int(font_size))
+            font = ImageFont.truetype(self.font_path, int(font_size * self.scale))
             (size_x, size_y), (offset_x, offset_y) = font.font.getsize(word)
             ascent, descent = font.getmetrics()
-            
+
             # Compute text bounding box
             min_x = offset_x
             max_x = size_x - offset_x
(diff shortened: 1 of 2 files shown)


jnothman:
Paste pics?

jojolebarjos:
Sure, here are rasterized versions of the SVG I attached:

- `to_image`: constitution
 - `to_svg` (rasterized with Inkscape): 
constitution svg
 - `to_image`: parrot_new
 - `to_svg` (rasterized with Inkscape): parrot_new svg

I also attach an example that use Segoe Script, to highlight differences between browsers:

 - `to_image`: 
airplane
 - `to_svg` (rasterized with Chrome): 
airplane svg chrome
 - `to_svg` (rasterized with Internet Explorer): 
airplane svg ie

jnothman:
Your PR has flake8 failures. Please add another commit rectifying these:
```

[scikit-ci] Executing: flake8
./wordcloud/wordcloud.py:733: [W293] blank line contains whitespace
    
^
./wordcloud/wordcloud.py:802: [W293] blank line contains whitespace
        
^
./wordcloud/wordcloud.py:822: [W293] blank line contains whitespace
            
^
./wordcloud/wordcloud.py:829: [E501] line too long (156 > 120 characters)
            # TODO some browser do not render glyphs the same way (e.g. Segoe Script in Chrome is different, while in Internet Explorer it matches to_image)
                                                                                                                        ^
./wordcloud/wordcloud.py:832: [F841] local variable 'min_y' is assigned to but never used
            min_y = ascent - size_y
            ^
```

wordcloud/wordcloud.py
```diff
@@ -728,6 +730,130 @@ def __array__(self):
 
     def to_html(self):
         raise NotImplementedError("FIXME!!!")
+    
+    def to_svg(self, embed_image=False):
+        """Export to SVG.
+
```
jnothman: please document `embed_image`

wordcloud/wordcloud.py
```diff
@@ -728,6 +730,130 @@ def __array__(self):
...
+            x *= self.scale
+            y *= self.scale
+            
+            # Get text metrics
+            font = ImageFont.truetype(self.font_path, int(font_size * self.scale))
+            (size_x, size_y), (offset_x, offset_y) = font.font.getsize(word)
+            ascent, descent = font.getmetrics()
+
+            # Compute text bounding box
+            # TODO some browser do not render glyphs the same way (e.g. Segoe Script in Chrome is different, while in Internet Explorer it matches to_image)
```
jnothman: yes, it looks like there are questions of ligatures and joining forms... not sure we want to fix this, but should document it.

wordcloud/wordcloud.py
```diff
@@ -728,6 +730,130 @@ def __array__(self):
...
+            else:
+                x += min_x
+                y += max_y
+                attributes['transform'] = 'translate({},{})'.format(x, y)
+            attributes['font-size'] = '{}'.format(font_size * self.scale)
+            attributes['style'] = 'fill:{}'.format(color)
+
+            # Create node
+            attributes = ' '.join('{}="{}"'.format(k, v) for k, v in attributes.items())
+            result.append('<text {}>{}</text>'.format(attributes, word))
```
jnothman: word needs to be XML-escaped.

jnothman:
I agree the masking of each word is not perfect, but it's decent...

jojolebarjos:
Thank you for your feedback, I really appreciate. I have fixed my code, according to your suggestions. Indeed, I'm not sure whether we can do anything about these differences in SVG renderers.

jnothman:
This lgtm

jojolebarjos:
I have added an option to embed the required font subset inside the SVG. This does solve the missing font problem, for instance when using the default wordcloud font. [fonttools](https://github.com/fonttools/fonttools) is required only if embed_font is true.

wordcloud/wordcloud.py
```diff
@@ -729,6 +732,244 @@ def __array__(self):
...
 
+    def to_svg(self, embed_font=False, optimize_embedded_font=True, embed_image=False):
+        """Export to SVG.
+
+        Font is assumed to be available to the SVG reader. Otherwise, text
+        coordinates may produce artifacts when rendered with replacement font.
+        It is also possible to include a subset of the original font in WOFF
+        format using `embed_font` (requires `fontTools`).
+
+        Note that some renderers do not handle glyphs the same way, and may
```
jnothman: But isn't the issue that PIL renders the font differently from some output renderers, and hence the masking for building the word cloud is only approximate? Is there a way to get the renderer, or the embedding to emulate the rendering used by PIL? Or is that getting too crazy?

jojolebarjos: After some investigations, my guess is that some renderers do not properly handle [complex text layouts](https://en.wikipedia.org/wiki/Complex_text_layout). Chrome and Inkscape seem to do this, while Internet Explorer doesn't.

There have been [some discussions on Pillow's repository](https://github.com/python-pillow/Pillow/issues/1089) to support this, which have been solved by using `raqm`. However, this requires many dependencies, and it is not included by default (see [Pillow's installation guide](https://pillow.readthedocs.io/en/stable/installation.html)). You can check if you Pillow installation has it:

```
from PIL import features
print(features.check("raqm"))
```

If it is the case, then the output will be similar to Chrome, which is probably the correct way to render text. If not, then it will use a simpler strategy (i.e. glyphs are rendered independently).

Here is the output of `to_image` when `raqm`is available:

bar_radm

I think we should not assume that `raqm` is available. And since other renderers are not consistent anyway, it does not make sense to force one of those "styles".

Still, to actually answer your question, it might be possible to drop these complex layouts by stripping some information when embedding the font. The `raqm` feature would therefore need to be explicitly disabled.

However, I believe it is a separate issue, as this has a limited impact on the SVG result. The goal of this PR is not to challenge the current design, which has worked well for years. But, we can indeed open an issue to keep track of this.

amueller:
Hm looks like python34 is failing but looks unrelated. Thanks a lot for the contribution. Good to merge?

wordcloud/wordcloud.py
jnothman: Thanks for this! Great analysis. Could you tweak the note to emphasise that the relationship between the generation and the online rendering is what might make it look awkward? And maybe to say that if raqm is installed, generation will accommodate complex text layouts handled by some renderers.

jnothman:
would it be worth adding at least a smoke test before merge, if not an image comparison test?

amueller:
Smoke test should be good. So far I have not done any image comparison tests, for better or worse, so I wouldn't ask for that.

wordcloud/wordcloud.py
jnothman: And personally, I think disabling complex layouts is not a good way to go, particularly with respect to Arabic and Indic writing systems, unless we adjust unicode handling to correspond.

jojolebarjos: I have updated the docstring. Indeed, we should avoid hacking into this.

jojolebarjos:
I added a small smoke test, which run `to_svg` and check that it is valid XML.

michaelsjackson:
It seems we (I) can not access this new svg export feature from the command line yet? (wordcloud_cli.py).  It would be cool if it could be added there as well.

It could work as follows:
```
--imagefile /path/image.png   would export as png
--imagefile /path/image.svg   would export as svg

```
Or anything similar which works in the end.

amueller:
@michaelsjackson I think that would be cool!

EidrianGM:
Hi, when will the method to_svg() uploaded to pip ?

amueller:
There's some issues with the build right now: #535 
wordcloud right now uses a build system written by @jcfr and I'm not sure what the issue is.
If I don't hear from him, I might redo the whole build process which might be some work. Any help trying to debug this is welcome.

jcfr:
I will have a look at #535

amueller:
Thanks!

thaoth58:
Thanks, guys!
Waiting for this feature in pip

michaelsjackson:
which repository has to_svg, if any?

amueller:
released as 1.7.0 https://pypi.org/project/wordcloud/

KonradHoeffner:
Is SVG export now possible from the command line? If yes, how?

amueller:
@KonradHoeffner looks like it hasn't been added to the CLI yet. PR welcome :)

KonradHoeffner:
@amueller: I tried to create a PR but I'm not a Python expert and couldn't figure out how to develop your tool quickly, as the README doesn't contain a section about it. Maybe you could a section about that to the README?

## Steps tried

- fork the project
- create a virtual environment with venv
- python -r requirements.txt
- python -r requirements-dev.txt

```
(venv) word_cloud$ python wordcloud/wordcloud_cli.py 

To execute the CLI, instead consider running:

  wordcloud_cli --help

or

  python -m wordcloud --help
```

At this point, I didn't know how to continue and didn't have the time to continue, sorry.

PR #525, opened 2020-02-07
Increase the small canvas size to avoid fallback to fail

okomestudio:
This is a bug fix in small canvas tests where the fallback behaviors are tested when only one (or zero) word fits within the canvas. The word placement has a random component, and `test_small_canvas` occasionally fails to find a placement for a word before its font size goes below the minimum font size allowed. 

This fix increases the small canvas size by one pixel to avoid that edge case to occur. The two small canvas tests also explicitly tests for the number of elements in `WordCloud.layout_`, which basically determines the fallback behavior.


amueller:
I haven't observed the failure but this seems to strengthen the tests either way.

Issue #451, opened 2018-11-09
Bug when using stdout as imagefile?


lucbouge:
#### Description
Error in PIL when writing to stdout

#### Steps/Code to Reproduce

Contents of file `words.txt`
> Hello!

Run `wordcount_cli`

> $ wordcloud_cli --text words.txt 
> Traceback (most recent call last):
>   File "/usr/local/bin/wordcloud_cli", line 11, in <module>
>     sys.exit(main())
>   File "/usr/local/lib/python3.7/site-packages/wordcloud/__main__.py", line 33, in main
>     wordcloud_cli_main(*wordcloud_cli_parse_args(sys.argv[1:]))
>   File "/usr/local/lib/python3.7/site-packages/wordcloud/wordcloud_cli.py", line 93, in main
>     image.save(imagefile, format='png', optimize=True)
>   File "/usr/local/lib/python3.7/site-packages/PIL/Image.py", line 1969, in save
>     save_handler(self, fp, filename)
>   File "/usr/local/lib/python3.7/site-packages/PIL/PngImagePlugin.py", line 734, in _save
>     fp.write(_MAGIC)
> TypeError: write() argument must be str, not bytes

However, it works fine with an explicit output file.
> wordcloud_cli --text words.txt --imagefile words.png

#### Actual Results
See above

#### Versions

> $ python
> Python 3.7.1 (default, Nov  6 2018, 18:45:35) 
> [Clang 10.0.0 (clang-1000.11.45.5)] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import platform; print(platform.platform())
> Darwin-17.7.0-x86_64-i386-64bit
> >>> import sys; print("Python", sys.version)
> Python 3.7.1 (default, Nov  6 2018, 18:45:35) 
> [Clang 10.0.0 (clang-1000.11.45.5)]
> >>> import numpy; print("NumPy", numpy.__version__)
> NumPy 1.15.4
> >>> import matplotlib; print("matplotlib", matplotlib.__version__)
> matplotlib 3.0.1
> >>> import wordcloud; print("wordcoud", wordcloud.__version__)
> wordcoud 1.5.0

amueller:
Thanks for the report. It's likely that that's not currently supported. PR welcome :)

Issue #445, opened 2018-10-31
TypeError: write() argument must be str, not bytes


jaraco:
#### Description

Running wordcloud with simple input of "foo" results in TypeError.

#### Steps/Code to Reproduce

Using rwt to download and run wordcloud, I get a TypeError.

```
$ echo foo | rwt wordcloud -- -m wordcloud
Collecting wordcloud
  Using cached https://files.pythonhosted.org/packages/c7/07/e43a7094a58e602e85a09494d9b99e7b5d71ca4789852287386e21e74c33/wordcloud-1.5.0-cp37-cp37m-macosx_10_6_x86_64.whl
Collecting numpy>=1.6.1 (from wordcloud)
  Using cached https://files.pythonhosted.org/packages/a8/e1/838e35e6f44e2bc19bf902e945c34ae730a33a8e346d6208bf7c4d751416/numpy-1.15.3-cp37-cp37m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
Collecting pillow (from wordcloud)
  Using cached https://files.pythonhosted.org/packages/99/c8/550f3416afe7b6726efc8a7f2249a38d6ae65c3514ef6c36bdc8485868b7/Pillow-5.3.0-cp37-cp37m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl
Installing collected packages: numpy, pillow, wordcloud
Successfully installed numpy-1.15.3 pillow-5.3.0 wordcloud-1.5.0
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-_9ebsdpk/wordcloud/__main__.py", line 37, in <module>
    main()
  File "/var/folders/c6/v7hnmq453xb6p2dbz1gqc6rr0000gn/T/rwt-_9ebsdpk/wordcloud/__main__.py", line 33, in main
    wordcloud_cli_main(*wordcloud_cli_parse_args(sys.argv[1:]))
[...]

PR #551, opened 2020-05-15
write image to stdout by default

guanzo:
Pretty simple commit. Fixes: #451, #445 

Reading this ticket helped a bit https://bugs.python.org/issue14156

Line 57 was taken from the official `argparse` lib, in order to toggle between `sys.stdout` and `sys.stdout.buffer`, based on the mode
https://github.com/python/cpython/blob/b47950f8ce4d06d6a7dd8e2c8c0c704d4972be01/Lib/argparse.py#L1191

Line 64 & 65 are to fix the error caused by line 111
`ValueError: binary mode doesn't take an encoding argument`

line 111 switches `argparse.FileType` to your custom `FileType`.

Examples:
```sh
# current usage
$ wordcloud_cli --text ./text.txt --imagefile ./test.png

# explicitly write to stdout
$ wordcloud_cli --text ./text.txt --imagefile - > ./test.png

# implicitly write to stdout
$ wordcloud_cli --text ./text.txt > ./test.png
```

### DIFF
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -140,7 +140,7 @@ def test_unicode_with_stopwords():
     assert u'\u304D' in args['stopwords']
 
 
-def test_cli_writes_image(tmpdir, tmp_text_file):
+def test_cli_writes_to_imagefile(tmpdir, tmp_text_file):
     # ensure writing works with all python versions
     tmp_image_file = tmpdir.join("word_cloud.png")
 
@@ -149,8 +149,19 @@ def test_cli_writes_image(tmpdir, tmp_text_file):
     args, text, image_file = cli.parse_args(['--text', str(tmp_text_file), '--imagefile', str(tmp_image_file)])
     cli.main(args, text, image_file)
 
-    # expecting image to be written
+    # expecting image to be written to imagefile
     assert tmp_image_file.size() > 0
+
+
+def test_cli_writes_to_stdout(tmp_text_file, capsysbinary):
+    tmp_text_file.write(b'some text')
+
+    args, text, image_file = cli.parse_args(['--text', str(tmp_text_file)])
+    cli.main(args, text, image_file)
+
+    # expecting image to be written to stdout
+    out, _ = capsysbinary.readouterr()
+    assert len(str(out)) > 0
 
 
 def test_cli_regexp(tmp_text_file):
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -162,7 +162,7 @@ def test_cli_writes_to_stdout(tmpdir, tmp_text_file):
     tmp_text_file.write(b'some text')
 
     originalBuffer = sys.stdout.buffer
-    sys.stdout.buffer = open(tmp_image_file, 'wb+')
+    sys.stdout.buffer = tmp_image_file.open('wb+')
 
     args, text, image_file = cli.parse_args(['--text', str(tmp_text_file)])
     cli.main(args, text, image_file)
--- a/test/test_wordcloud_cli.py
+++ b/test/test_wordcloud_cli.py
@@ -1,6 +1,7 @@
 import argparse
 import os
 import subprocess
+import sys
 from collections import namedtuple
 
 import wordcloud as wc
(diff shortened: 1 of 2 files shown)


amueller:
Great, can you please add a test?

guanzo:
I've zero experience in testing python. How would I test that the `wordcloud` command writes to stdout? Do i actually need to run the wordcloud cli and check the stdout? Or can I mock it somehow.

I'm looking at `test_wordcloud_cli.py`. Maybe copy and modify `test_cli_as_executable`?

guanzo:
Trying but running into cryptic errors. Assist?

```py
def test_cli_writes_to_stdout(tmp_text_file, capsys):
    tmp_text_file.write(b'some text')

    args, text, image_file = cli.parse_args(['--text', str(tmp_text_file)])
    cli.main(args, text, image_file)

    captured = capsys.readouterr()
    # expecting image to be written to stdout
    assert len(captured.out) > 0
```

throws

```py
 File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\capture.py", line 329, in _install_capture_fixture_on_item
    yield fixture
  File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\capture.py", line 356, in close
    out, err = self._capture.pop_outerr_to_orig()
  File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\capture.py", line 481, in pop_outerr_to_orig
    out, err = self.readouterr()
  File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\capture.py", line 523, in readouterr
    self.out.snap() if self.out is not None else "",
  File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\capture.py", line 660, in snap
    res = self.tmpfile.getvalue()
  File "c:\users\eric\appdata\local\programs\python\python38-32\lib\site-packages\_pytest\compat.py", line 379, in getvalue
    return self.buffer.getvalue().decode("UTF-8")
ValueError: I/O operation on closed file.
```

what i think is happening is `with imagefile` closes stdout after `image.save`
so by the time `capsys.readouterr()` runs, stdout is closed, and the error is thrown
i've tried moving `capsys.readouterr()` to before the cli.main call, but same error.

guanzo:
Ok looks like my intuition was correct. Looks context manager + stdout is a common problem. 
https://stackoverflow.com/questions/17602878/how-to-handle-both-with-open-and-sys-stdout-nicely

The answers have way too much code for my liking. This PR is small, I don't want to add 50 lines just to handle this, so I replace `with` with `try/finally`.

guanzo:
> E       capsysbinary is only supported on Python 3
 
zzzzzzzzzzzzzzzzzzzzzzzz. what do.

guanzo:
a bittersweet ending. plz merge thx

test/test_wordcloud_cli.py
```diff
@@ -149,7 +150,26 @@ def test_cli_writes_image(tmpdir, tmp_text_file):
...
+
+
+# capsysbinary should be used here, but it's not supported in python 2.
+def test_cli_writes_to_stdout(tmpdir, tmp_text_file):
+    # ensure writing works with all python versions
+    tmp_image_file = tmpdir.join("word_cloud.png")
+
+    tmp_text_file.write(b'some text')
+
+    originalBuffer = sys.stdout.buffer
```
amueller, line 164: You could have used ``patch``: https://www.geeksforgeeks.org/python-testing-output-to-stdout/

but this solution is also fine, I think.

amueller:
Great, thank you! Why bittersweet? looks good!

guanzo:
Because I thought this PR would be really easy, but it ended up taking a lot more hours of my time than i wanted it to 🙁. I'm just grumpy, thanks for merge. :)

amueller:
PRs usually do, that's murphy's law of PRs (or Hofstadter's law maybe) ;)

test/test_wordcloud_cli.py
```diff
@@ -149,7 +150,26 @@ def test_cli_writes_image(tmpdir, tmp_text_file):
...
+    sys.stdout.buffer = tmp_image_file.open('wb+')
```
Mic92, line 165: This actually breaks with python38 and newer because `sys.stdout.buffer` is now read-only.

PR #315, opened 2017-10-27
Arabic example for word_cloud

caleighm:
Here's a little Arabic example - I noticed a couple issues that folks had with Arabic, and in one of them you responded to say it would be helpful to have an example in the documentation. So, here's one!

PR #554, opened 2020-05-15
Add Arabic wordcloud example (in a readme file) See #315

iamaziz:


AMR-KELEG:
Hi @iamaziz ,
Thanks for taking the initiative to finally add the example to the package.
I will share my thoughts regarding this example.

IMHO, we should work on merging the original pull request and give credit to the original author (https://github.com/amueller/word_cloud/pull/315/).
OTOH, the current status of the original PR indicates that it needs some minor bug fixes (Renaming filename, adding the font that you have used "which is better IMO" and modifying the long comment messages used there).

For your script:
- You have added some lines for counting the words which is the only difference to the original script (https://github.com/amueller/word_cloud/blob/cb926bd718f654bd588b2f29c5fcd482961dd01d/examples/arabic.py)
This is the output of the original script after applying bug fixes.
arabic_example
Will using a `Counter` yield better results?
- A README is always nice to have but I am think that the whole script should be moved to a separate `.py` file (which is the convention that is currently used in the repository).

iamaziz:
Yea definitely agreed all the credit goes to the original PR (and its comments), it's linked in the title of this PR. I just thought it would be nicer (at least) to have an example for a quick how-to reference so people can see and adopt (even as a readme file, after failing the CI https://github.com/amueller/word_cloud/pull/553 with .py file). Instead of muddling in the comments of a three-year-old PR just to figure out a simple step.

This PR is probably not the better direction, please feel free to take over both PRs or close this one :] and let me know if can help

iamaziz:
I felt guilty for the state of this PR :/ @AMR-KELEG perhaps it makes sense if we close it, for now, to follow up with the original PR and the example wrapper there.

Issue #558, opened 2020-05-26
Stopword filtering seems off with version 1.7.0


freeIsa:
#### Description
With version 1.7.0, wordclouds still include words I explicitly added to my stopwords list. Same code works fine with 1.6.0.

#### Steps/Code to Reproduce
Example:
```python
    add_stopwords = ['ciao', 'ovviamente']
    stopwords = open(INPUT_FOLDER + STOPWORDS_FILE, encoding="utf-8").readlines()
    stopwords += add_stopwords
    stopwords = set([s.rstrip() for s in stopwords])

    # generate wordcloud
    wordcloud = WordCloud(stopwords=stopwords, width=1200, height=600).generate(all_comments)
```
#### Versions
macOS-10.15.4-x86_64-i386-64bit
Python 3.8.3 (v3.8.3:6f8c8320e9, May 13 2020, 16:29:34) 
[Clang 6.0 (clang-600.0.57)]
NumPy 1.18.4
matplotlib 3.2.1
wordcoud 1.7.0

amueller:
Please provide [self-contained example code](https://stackoverflow.com/help/mcve), including imports and data (if possible), so that other contributors can just run it and reproduce your issue. Ideally your example code should be minimal.

I suspect this is a consequence of #557 and the stop words are included in collocations.

freeIsa:
> Please provide [self-contained example code](https://stackoverflow.com/help/mcve), including imports and data (if possible), so that other contributors can just run it and reproduce your issue. Ideally your example code should be minimal.

I made this snippet, I hope this helps.

``` python
    stopwords = ['the', 'a', 'of', 'and', 'it', 'Lorem', 'Ipsum']
    all_text = 'Lorem Ipsum is simply dummy text of the printing and typesetting industry. \
                    Lorem Ipsum has been the industry\'s standard dummy text ever since the 1500s, \
                    when an unknown printer took a galley of type and scrambled it to make a type \
                    specimen book. It has survived not only five centuries, but also the leap \
                    into electronic typesetting, remaining essentially unchanged. \
                    It was popularised in the 1960s with the release of Letraset sheets containing \
                    Lorem Ipsum passages, and more recently with desktop publishing software \
                    Lorem Ipsum is simply dummy text of the printing and typesetting industry. \
                    Lorem Ipsum has been the industry\'s standard dummy text ever since the 1500s, \
                    when an unknown printer took a galley of type and scrambled it to make a type \
[...]

PR #561, opened 2020-06-04
Handle case in collocations

freeIsa:
As discussed in issue #558 I made a change in function unigrams_and_bigrams so that words are lower cased before being checked against stopword list.
Formatting and test passages were ok.

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -48,6 +48,17 @@ thank you very much
 thanks
 """
 
+STOPWORDED_COLLOCATIONS_UPPERCASE = """
+Thank you very much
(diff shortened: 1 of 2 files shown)


amueller:
Thank you, that looks good! Please add a [non-regression test](https://en.wikipedia.org/wiki/Non-regression_testing) that would fail at master but pass in this PR. You can add it to the ``test_wordcloud.py`` file and add a new function there that starts with ``test_``. Or if there's a similar related test already, you can add a check to that test.

freeIsa:
Test added!

amueller:
Thanks!

PR #564, opened 2020-06-11
fix #542 plurals issue

### LINKED ISSUE #542: Normalise Plurals feature causes problems when text includes spelling mistakes
#### Description
If the text contains spelling mistakes of the form where the ending 's' of a word is skipped, then the normalise_plural feature removes the original word and keeps the spelling mistake

#### Steps/Code to Reproduce

Example:
```python
from wordcloud import WordCloud
mpl.use('Agg')
import matplotlib.pyplot as plt
from wordcloud import WordCloud
a = 'this is a wordcloud show test'
wc = WordCloud().process_text('viru virus')
print(wc)
```

#### Expected Results
Ideally, one would expect the word "virus" to be in the resulting dictionary, wc['virus'] = 1.

#### Actual Results
But both the tokens are counted as 'viru'

#### Versions
Linux-4.19.104+-x86_64-with-Ubuntu-18.04-bionic
Python 3.6.9 (default, Nov  7 2019, 10:44:02) 
[GCC 8.3.0]
NumPy 1.18.3
matplotlib 3.2.1
wordcoud 1.5.0

#### Suggested Solution
If the user has the knowledge about which words should not be normalized for plurality, they should be able to provide that to word cloud

HACore:
Hi.
We solved #542 the plurals issue by utilizing the pyenchant package. It's a way that doesn't affect execution time as much as you worry about. If there is no problem with this after the review, please merge it to the project.
You can use `pip install pyenchant`

### DIFF
--- a/wordcloud/tokenization.py
+++ b/wordcloud/tokenization.py
@@ -3,6 +3,7 @@ from itertools import tee
 from operator import itemgetter
 from collections import defaultdict
 from math import log
+from enchant import Dict
 
 
 def l(k, n, x):  # noqa: E741, E743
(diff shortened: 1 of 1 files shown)


amueller:
Wouldn't it be easier to just spellcheck the text before passing it to wordcloud?

PR #574, opened 2020-07-23
Added a switch to allow turning off color random value

daqieq:
Needed a way to set the default color used by GroupedColorFunc to always be the same color shade (the value part of the HSV color). Added an optional boolean switch to allow this functionality while not changing the original random_state functionality.

Basic code change, no formatting issues with flake8, and the 4 failed test cases did not appear to be related to my change with with stopwords and the cli which I did not change:

Results (13.31s):
      70 passed
       4 failed
         - test/test_wordcloud.py:79 test_collocation_stopwords
         - test/test_wordcloud.py:90 test_collocation_stopwords_uppercase
         - test/test_wordcloud.py:458 test_plural_stopwords
         - test/test_wordcloud_cli.py:158 test_cli_writes_to_stdout

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -114,13 +114,13 @@ class colormap_color_func(object):
         return "rgb({:.0f}, {:.0f}, {:.0f})".format(r, g, b)
 
 
-def get_single_color_func(color):
+def get_single_color_func(color, random_value=True):
     """Create a color function which returns a single hue and saturation with.
     different values (HSV). Accepted values are color strings as usable by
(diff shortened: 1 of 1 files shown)


amueller:
Thanks for the PR.
Can't you just use ``lambda x: colorvalue`` as ``color_func``?

daqieq:
Yes, that works for my use case:
`self.default_color_func = lambda *args, **kwargs: "red"`
I'll cancel/close this PR because I can modify `GroupedColorFunc` with this to get the result I want. Thanks for the review!

Issue #576, opened 2020-07-30
Examples don't run on some platforms


GadgetSteve:
#### Description

On Windows 10 at least some of the examples do not run giving encoding errors.

#### Steps/Code to Reproduce
```
cd examples
py parrot.py
# UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 32995: character maps to <undefined>
py emoji.py
# UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 1214: character maps to <undefined>

#### Expected Results
<!-- Example: No error is thrown. Please paste or describe the expected results.-->
Display of word cloud(s)

#### Actual Results
<!-- Please paste or specifically describe the actual output or traceback. -->
> py parrot.py
Traceback (most recent call last):
  File "parrot.py", line 23, in <module>
    text = open(os.path.join(d, 'wiki_rainbow.txt')).read()
  File "C:\Python38_64\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 32995: character maps to <undefined>

> py emoji.py
Traceback (most recent call last):
  File "emoji.py", line 27, in <module>
    text = io.open(path.join(d, 'happy-emoji.txt')).read()
  File "C:\Python38_64\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 1214: character maps to <undefined>

#### Versions

Windows-10-10.0.18362-SP0
Python 3.8.5 (tags/v3.8.5:580fbb0, Jul 20 2020, 15:57:54) [MSC v.1924 64 bit (AMD64)]
NumPy 1.19.1
matplotlib 3.3.0
wordcoud 1.7.0

<!-- Thanks for contributing! -->

GadgetSteve:
This is due to the encoding not being specified on opening the file(s) adding `, encoding='utf-8')` to the open calls fixes these issues.

PR #577, opened 2020-07-30
Add encoding to two examples

GadgetSteve:
Address #576

### DIFF
--- a/examples/emoji.py
+++ b/examples/emoji.py
@@ -24,7 +24,7 @@ from wordcloud import WordCloud
 d = path.dirname(__file__) if "__file__" in locals() else os.getcwd()
 
 # It is important to use io.open to correctly load the file as UTF-8
-text = io.open(path.join(d, 'happy-emoji.txt')).read()
+text = io.open(path.join(d, 'happy-emoji.txt'), encoding='utf-8').read()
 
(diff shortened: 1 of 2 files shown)


GadgetSteve:
Also fixed Examples/wordcloud_cn.py

examples/wordcloud_cn.py
```diff
@@ -12,9 +12,14 @@
 at the same time using wordcloud with jieba very convenient
 """
 
-import jieba
-jieba.enable_parallel(4)
-# Setting up parallel processes :4 ,but unable to run on Windows
+import sys
+try:
```
amueller, line 16: you can do an ``if __name__ == "__main__"`` that should fix that, I think?

amueller:
Thanks for the PR! I restarted the CI that had some random issue,
For multiprocessing in windows, you can put the whole example into an ``if __name__ == __main__`` block or put it in a function and then protect a call to the function that way.

GadgetSteve:
No – the trick is not to try to run parallel on Windows (at the moment) if __name__ == “__main__” will be as true on Windows as on any other platform.

amueller:
See https://stackoverflow.com/questions/20222534/python-multiprocessing-on-windows-if-name-main for example

Issue #495, opened 2019-08-29
Single words are not being targeted: Regex Issue


dipanghosh:
#### Description
I was after a special application, so needed to generate word cloud from single letter or numbers. It turns out the current implementation has a bug that does not allow this. 
#### Steps/Code to Reproduce
```
text = "a b c 1 2 3"
wc = WordCloud(repeat = False, width=250, height=250, include_numbers=True, min_word_length=1, scale)
wc.generate(text)
```
#### Expected Results
WordCloud object

#### Actual Results
ValueError: We need at least 1 word to plot a word cloud, got 0.

#### The Bug
I looked through the source and tracked down the problem. The culprit is the default regex defined in the process_text method. The regex is `\w[\w']+` which stands for any word, followed by any other word. So any single character gets ignored. This may be an intended feature, but at least if the min_word_length is set to 1 or 0, this should be taken into account.

#### Potential Fix
Just changing the regex to `\w+` have resolved the problem for me.

amueller:
PR welcome, indeed we should probably special-case this for ``min_word_length=0`` and change the default regex in that case.

PR #585, opened 2020-09-11
regex change to address issue #495

jamesgagg:
Changed r"\w[\w']+" to r"\w[\w']*" so that single words are matched too, as per issue #495.  
Note: the user will need to be mindful that single letters words such as 'a' and 'i' will not appear if the default stopword list is used.

### DIFF
--- a/test/test_wordcloud.py
+++ b/test/test_wordcloud.py
@@ -327,6 +327,20 @@ def test_process_text():
     assert isinstance(result, dict)
 
 
+def test_process_text_default_patterns():
+    wc = WordCloud(stopwords=set(), include_numbers=True, min_word_length=2)
+    words = wc.process_text(THIS)
+
+    wc2 = WordCloud(stopwords=set(), include_numbers=True, min_word_length=1)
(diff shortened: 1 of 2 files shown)


amueller:
Thanks for the PR! I wanted to change the regexp just in the case when users use ``min_word_length=1``. Otherwise I see no reason to change the default, the user can always just pass the regexp they want.

jamesgagg:
Thank you for the feedback, the PR has now been updated accordingly.

amueller:
Great, thank you. Can you please add a test as well?

jamesgagg:
Yes, no problem, I should have some time at the end of the week.

Also, thanks for your patience, as you can probably tell this is my fist PR.

amueller:
All good, actually I couldn't tell :) Sorry my replies are a bit slow these days, I have a lot going on.

jamesgagg:
Thanks and no worries re the late replies, same for me too. Test has now been added.

amueller:
great, thank you!

PR #600, opened 2020-09-27
To html fix

nishd8:
HTML implementation of word cloud in to_html() function


amueller:
Thanks for the PR! It's probably easier to just call the ``to_svg`` method instead. The ``to_html`` method should probably be removed.
Also, you don't need to open new PRs every time you update the branch, you can just push into the same branch again to update the PR.

PR #608, opened 2020-10-30
convert np.float to float for positions

spacewaterbear:
convert np.float to float (native python). It cannot be jsonify otherwise


amueller:
Hey, thanks for the PR. What is your goal?

amueller:
Sorry for the slow reply. You could just do this conversion on your side, right? It should be pretty easy to iterate over ``.layout_`` and do the casting.

PR #616, opened 2020-12-30
test_wordcloud_cli: fix stdout mock for python3.8+

Mic92:


Mic92:
This breaks on python2.7 of course. But I refuse to support insecure python versions.

amueller:
Thanks and sorry for the slow reply! I need to fix up CI but this seems good.

Issue #205, opened 2016-12-08
wc object not updating width and height if using mask


Xunius:
Hi amueller,
Thanks for the hard work.
I was playing around with mask and re-coloring, and noticed that if I want to resize my coloring image to the same size as the generated text image, I had to query the mask's size rather than the wc object, e.g.

```
wc = WordCloud(background_color="white", max_words=2000, mask=alice_mask,
               stopwords=stopwords,width=width,height=height,random_state=100)
wc.generate(text)
newshape=wc.mask.shape
```
As stated in the API doc, `width` and `height` are ignored in the constructor if `mask` is given, but if I supply them, they will be stored in the `wc` obj. So I wonder will it be better to update these 2 properties in the constructor?

amueller:
Sorry, I'm not sure what you mean by "query the mask size rather than the wc object". You mean that ``wc.width`` and ``wc.height`` are invalid? Yeah, that's true. We could set them in the constructor. I didn't really see a reason to do that, but I guess there's no harm.
The reliable way to get the shape of ``wc`` *after* you generated it is ``wc.to_array().shape``.

Xunius:
Yes I meant `wc.width` and `wc.height` should be set to the size of the mask array, if a mask is given. Otherwise they will be, for instance (600,400) if they are given so in the constructor, even though the mask is actually (900,900). The reason is obvious: for the sake of consistency and the ease of use, the user doesn't have to read the API to figure out what they should query, but instead, the `width` and `height` names should be intuitive enough to infer these are the properties they want. Are they meant to infer the array size, or something else? what's the point of setting these 2 attributes in the first place?

amueller:
Well they are meant to give the image size if you don't use a mask.
We could not set them if a mask is given or set them to the mask size. Feel free to submit a PR.

richardfat7:
I found a bug related to this issue.

In wordcloud.py function generate_from_frequencies line [447](https://github.com/amueller/word_cloud/blob/b6f48e108224f84b0b1659cea8558c86ccfc9898/wordcloud/wordcloud.py#L447)
when it try to figure out the max font size from drawing first 2 words, it set max font size temporary as `self.height`, while it will work normally if we use `width` and `height` to specify the size, it could not work when only `mask` is supplied. The `self.height` will remain as the default 200 without respecting to the mask size which is the one used in drawing. If largest font size at the end is less than 200, then this will not be a problem, maybe that's why it is not easy to discover.

Code to reproduce the issue:
```
[...]

PR #619, opened 2021-01-21
fix the max font size not respecting mask height and add a test on it

richardfat7:
Address #205 
flake 8 result is empty
pytest result have some error on test_wordcloud_cli, but I guess that shd have been fixed in other PR


amueller:
thanks! sorry for the slow reply. and yes, CI needs some love :-/

PR #631, opened 2021-03-21
Feature: to_html implemented

theoYe:
1. I guess there are only two type of html can be exported as, which are
- canvas (pixels graph)
- svg (vector graph)

luckily , although they are two different type of graph , canvas is transformed just the same as svg.so thanks to @jojolebarjos , I can finish part of "to_canvas" quickly with reference to his to_svg method.

Here are some differences I've found bettween canvas and svg:
1.  canvas's performance in html is much better than svg
2.  canvas won't load the font automatically, so I need to write some javascript code in my html template and put a div tag in int to load the font

2. support font family will fix the texts overlap problem in to_svg method

when I tested the to_svg method, I found there are some text overllaped together , but after I add a font family attribute , the problem is gone

3.. some problems
- the font load mechanism I am using might have some compability issues, but I am pretty sure it worked well in chrome and miscrosoft edge 
- there's a a lot repeat code fragment in to_svg and to_canvas methods , may need a refactoring here

### DIFF
--- a/examples/a_new_hope_to_html.py
+++ b/examples/a_new_hope_to_html.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+"""
+Using custom colors
+===================
+
+Using the recolor method and custom coloring functions.
+"""
+
+import numpy as np
+from PIL import Image
+from os import path
+import matplotlib.pyplot as plt
+import os
+import random
+
+from wordcloud import WordCloud, STOPWORDS
+
+
+def grey_color_func(word, font_size, position, orientation, random_state=None,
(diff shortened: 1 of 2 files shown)


theoYe:
why can't I pass those checks ........ 😭

amueller:
hey! Thanks for the PR. I thought I had remove the method, as I figured the SVG export would be enough.
Do you feel like having html output in addition to SVG is useful? What would be the use cases?

amueller:
Also, I think, CI is just broken right now and it's not your fault :(

PR #695, opened 2022-10-29
Compiled query_integral_image.pyx w/ newer cython

TheGhostOfInky:
Fixes to compile under Python 3.11


amueller:
Fixed in #697, sorry about the delay.

PR #699, opened 2022-11-10
Create stopwords_sp.txt

diegotco:
This file has Spanish words and should be helpful for these developers


diegotco:
Is there any blocker to merge this? Please let me know @amueller

amueller:
Hey! sorry for the delay, I somehow missed the PR. I think I'd rather not expand the stopword list. Including the English one is already a bit arbitrary because there is really no such thing as a general purpose stopword list. For example your list seems to include "last" and "new", I don't know why these would be considered stopwords, or in which context. If we start adding other languages, I feel this will easily get out of hand. Also, I can't maintain the list because I can't speak Spanish.
To do a proper Spanish word cloud, you probably want to use a stemmer so you can aggregate the different declinations anyway.

PR #703, opened 2022-12-06
Solution to Issue 274

Prashiksha:
Solution to ZeroDivisionError
The file named issue274_resolved.py has a sample code showing how the word "funny" and "not funny" can be incorporated in wordcloud simultaneously.


amueller:
Sorry for the slow response. I'm not entirely sure if adding this example will help. It'll be great to provide a better error that's more actionably, but I haven't thought about how to best do that.

PR #706, opened 2022-12-24
[MNT] - Update README for list of tested python versions

TomDonoghue:
The README is a bit out of date with the current tested versions, and in the current version implies that the module still is tested on / supports Python2, which I think is no longer the case (?). This update lists the current set of tested Python versions, drawing from the Github actions file.

In addition, there is a minor update to the install section of the README to note that matplotlib is now a required dependency (it previously stated it was optional, but I believe this is outdated).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ word_cloud
 A little word cloud generator in Python. Read more about it on the [blog
 post][blog-post] or the [website][website].
 
-The code is tested against Python 2.7, 3.4, 3.5, 3.6 and 3.7.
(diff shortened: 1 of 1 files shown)


amueller:
Thank you! Looks good.

amueller:
Can you please make sure your original branch is up-to-date?

TomDonoghue:
My fork says it's up to date with `amueller:master`

amueller:
thank you for checking! I'm confused why the test are failing then. But this PR is clearly not the cause.

PR #685, opened 2022-07-24
Fix some deprecated

teriterance:
This pull request remove some warning for deprecated variables

`lib/python3.8/site-packages/wordcloud/wordcloud.py:501: DeprecationWarning: textsize is deprecated and will be removed in Pillow 10 (2023-07-01). Use textbbox or textlength instead.`
I used textlength in place of textsize

np.int is a deprecated, replaced by int. 

and ROTATE_90 replaced by Transpose.ROTATE_90

PR #711, opened 2023-02-25
Fix test failures and deprecated Pillow functions

dataymeric:
Fixes #709 : It appeared tests were failing because of now deprecated `np.int` used in the creation of masks. It has been changed to `int` to avoid failures.

Fixes #685 #705 : as mentionned, some Pillow functions were soon deprecated, which were `textsize()` and `getsize()`. They have been changed to their counterpart `textbbox()` and `getbbox()`.

I also looked at the deprecated `plt.cm.get_cmap()` but as of now, I am not sure how to deal with it. We should use `matplotlib.colormaps[name]` or `matplotlib.colormaps.get_cmap(obj)`, but the first only deals with strings and the latter is not available in Python 3.7.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,7 +31,7 @@ jobs:
     name: "Build aarch64 ${{ matrix.pyver }}"
     strategy:
       matrix:
-        pyver: [cp37-cp37m, cp38-cp38, cp39-cp39, cp310-cp310]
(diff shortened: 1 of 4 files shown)


stevenaivarsson:
Hi @dataymeric  How far until this branch is merged? Thanks You

amueller:
Sorry for the slow response, I pushed to the branch to trigger CI, which should be fine, now?

PR #721, opened 2023-04-28
Fix matplotlib deprecation warning and adds some minor improvements

amenezes:
- Fix deprecated matplotlib function
``` txt
test/test_wordcloud.py: 46 warnings
  wordcloud/wordcloud.py:106: MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
    self.colormap = plt.cm.get_cmap(colormap)
```
- Added metadata classifiers
- Improve regex performance using [re.compile](https://docs.python.org/3/library/re.html#re.compile)

### DIFF
--- a/wordcloud/wordcloud.py
+++ b/wordcloud/wordcloud.py
@@ -103,7 +103,7 @@ class colormap_color_func(object):
     """
     def __init__(self, colormap):
         import matplotlib.pyplot as plt
-        self.colormap = plt.cm.get_cmap(colormap)
+        self.colormap = plt.colormaps.get_cmap(colormap)
 
     def __call__(self, word, font_size, position, orientation,
                  random_state=None, **kwargs):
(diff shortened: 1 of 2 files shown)


wordcloud/wordcloud.py
```diff
@@ -576,13 +576,14 @@ def process_text(self, text):
 
         flags = (re.UNICODE if sys.version < '3' and type(text) is unicode  # noqa: F821
                  else 0)
-        pattern = r"\w[\w']*" if self.min_word_length <= 1 else r"\w[\w']+"
-        regexp = self.regexp if self.regexp is not None else pattern
+        if self.regexp is not None:
```
amueller, line 579: Thanks for the PR!
This call to compile actually doesn't change anything, as the object is only used once, right?

amenezes: @amueller,

Really sorry for the long delay. Yes there's no change on the workflow. The ideia is to simplify.

All the tests have passed locally, do you think it's useful to include any more?

amueller: No worries, I'm only very sporadically on this project these days. How does this change about the re simplify, the code gets longer, right?

QuLogic:
Just a friendly ping here; Matplotlib 3.9 will be out soon and this deprecation will be a hard error.

PR #732, opened 2023-08-12
MAINT Use unittest.mock instead of mock

davide125:
The `mock` package has been deprecated for a while, use the builtin `unittest.mock` instead.


thomasjpfan (approved):
Thank you for the PR! LGTM

PR #745, opened 2023-12-05
Migrate to pyproject.toml

thomasjpfan:
This PR migrates the build system to use `pyproject.toml`.

This PR also replaces `versioneer.py` with `setuptools_scm`. It works the same way, where the version is generated based on the git tag. During wheel or sdist building, `setuptools_scm` writes the version to a `_version.py` file.


thomasjpfan:
I tested wheel building with this configuration on my fork and it all works: https://github.com/thomasjpfan/word_cloud/actions/runs/7109479262

amueller:
thank you! lgtm

PR #747, opened 2023-12-06
Update Python versions in CI and README

thomasjpfan:
This PR updates the Python versions in the README and the build system.


thomasjpfan:
With https://github.com/amueller/word_cloud/pull/745, looks like cibuildwheel can figure out the require python version from the `pyproject.toml`. So there is no need to configure cibuldwheel.

Issue #791, opened 2025-03-27
Error building wheel due to license in pyproject.toml not confirming to PEP 639


jhamon:
I see this when trying to install:

```  
Building wheel for wordcloud (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for wordcloud (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [44 lines of output]
      /tmp/pip-build-env-ckk8qifs/overlay/lib/python3.12/site-packages/setuptools/config/_apply_pyprojecttoml.py:82: SetuptoolsDeprecationWarning: `project.license` as a TOML table is deprecated
      !!
      
              ********************************************************************************
              Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).
      
              By 2026-Feb-18, you need to update your project and remove deprecated calls
              or your builds will no longer be supported.
      
              See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
              ********************************************************************************
      
      !!
```

PR #792, opened 2025-03-27
Update license metadata for PEP 639

jhamon:
## Problem

I think this will resolve the error reported in  #791 

## Solution

Under [PEP 639](https://peps.python.org/pep-0639/), license metadata should be expressed using [SPDX license identifiers](https://spdx.org/licenses/).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ authors = [{ name = "Andreas Mueller", email = "t3kcit+wordcloud@gmail.com" }]
 description = "A little word cloud generator"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.7"
-license = { text = "MIT License" }
+license = "MIT"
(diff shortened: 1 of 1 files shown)


thomasjpfan (dismissed):
LGTM

pyproject.toml
```diff
@@ -8,7 +8,7 @@ authors = [{ name = "Andreas Mueller", email = "t3kcit+wordcloud@gmail.com" }]
 description = "A little word cloud generator"
 readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.7"
-license = { text = "MIT License" }
+license = "MIT"
```
thomasjpfan, line 11: Looks like this new spec requires a newer version of `setuptools`, which does not support Python 3.7.

I think it's time to drop 3.7, but for the time being, I think we can put an upper bound on, `setuptools<=68.0.0`, in `[build-system]`.

thomasjpfan (approved):

PR #804, opened 2025-11-30
Update cibuildwheel version to v3.3.0 to get python-3.14 wheels

stonebig:
I'm not sure it's the way, but I don't see another way. Having Python-3.14 wheels, even just for the non free-threading version,  would be great


thomasjpfan (approved):
Thank you for the PR! LGTM

